PR #8, opened 2024-05-02
Bunch of changes to lnd.py

starbackr-dev:
1. Added Nostr to post logs every 5 mins
2. Reduce calculated risk score if the other side recovered from issues
3. increase sleep time to 60 before it checks if the other side posted payments
4. added .env file to move logs files path and nostr private keys
5. Changes to log json format to make it easier to post nostr events.


toneloc:
All these changes seems out of scope for main stable channels branch, which (hopefully) will be used by others with different designs.

PR #9, opened 2024-05-17
cln-plugin: make it more universally useable + tests + CI

daywalker90:
Hi,
interesting project you have here. I saw that you had no tests for you cln-plugin and since i'm a cln plugin dev myself i thought i'd help you get started on that. This PR contains:

- Some fixes to make the plugin less dependable on hard coded stuff like the log file paths, the python env path
- Removed the need to pass in the rpc path, we can get it from the plugin
- I added the execution bit to the plugin file so not everyone has to do it again
- Also replaced all `print`'s with `plugin.log` since `print` should not be used in cln plugins (or atleast you should not write non json-rpc line to stdout)
- Boolean options can be passed in as `str` or `bool` so i handled both cases
- Added a basic startup test that runs the `check_stables` function once (i've added a `dev_check_stables` rpc method to the plugin for this) and triggers the `handle_coin_movement` event
- Added CI that will run the tests for the cln plugin on every PR and on commits to main (you can test more cln versions by copying `main_v24.02.yml` and editing the versions)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,140 @@
+name: CI
+
+# Cancel duplicate jobs
+concurrency:
+    group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+    cancel-in-progress: true
+
+on:
+    workflow_call:
+        inputs:
+          cln-version:
+            required: true
+            type: string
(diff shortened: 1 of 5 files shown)


toneloc (approved):

toneloc:
Looks great.

PR #26, opened 2026-04-03
fix: ensure data directory is created on first run

geeekyvishal:
Fixes a panic on first run by ensuring the user data directory is created before saving the seed phrase.

Tested by deleting the directory and verifying successful initialization.


toneloc:
Gonna close this. Don't want the .gradle cache files / Android checksums into the repo, uses .expect() (hard panic) instead of error handling.    

Superseded by PR #29 which does the same thing

PR #27, opened 2026-04-03
fix: reconcile_incoming does not reset backing_sats after payment received

Sandipmandal25:
### Description

`reconcile_incoming` was not resetting `backing_sats` after a stability payment is received.

When a stability payment occurs (i.e., price drops and the LSP compensates the user), the `backing_sats` must be updated to reflect the new equilibrium:

```
backing_sats = expected_usd / price * 1e8
```

Without this reset, subsequent stability checks compute drift against an outdated baseline, leading to compounding errors across multiple payment cycles.

Two existing tests in the repository were intended to catch this issue but were failing silently. This change corrects the behavior and ensures both tests pass.

### Changes

* `src/stable.rs`

  * Reset `backing_sats` before calling `recompute_native`
  * Add guard to skip update when price is zero

* `ios/StableChannels/StableChannels/Services/StabilityService.swift`

  * Apply equivalent fix to reset `backing_sats` in Swift implementation

### DIFF
--- a/ios/StableChannels/StableChannels/Services/StabilityService.swift
+++ b/ios/StableChannels/StableChannels/Services/StabilityService.swift
@@ -88,8 +88,12 @@ enum StabilityService {
         sc.nativeChannelBTC = Bitcoin(sats: nativeSats)
     }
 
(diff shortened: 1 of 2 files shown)


toneloc:
Fixed this earlier

PR #30, opened 2026-04-05
feat(android): use build variants for network/Esplora configuration

DevAnuragT:
feat(android): add build variant-based network configuration

Replaces hardcoded Network.BITCOIN and Esplora URLs with 
BuildConfig fields driven by build type:

- debug builds → Signet (mutinynet.com)  
- release builds → Mainnet (blockstream.info)

This allows contributors to test safely without touching 
real funds. No UI changes, no user-facing impact.

Updated all hardcoded references across:
- Constants.kt
- AppState.kt (3 call sites)
- StabilityProcessingService.kt
- SettingsScreen.kt

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -23,6 +23,9 @@ android {
             buildConfigField("String", "NETWORK", "\"signet\"")
             buildConfigField("String", "ESPLORA_URL", "\"https://mutinynet.com/api\"")
             buildConfigField("String", "FALLBACK_ESPLORA_URL", "\"https://mutinynet.com/api\"")
+            buildConfigField("String", "LSP_PUBKEY", "\"\"") // Signet LSP placeholder
+            buildConfigField("String", "LSP_ADDRESS", "\"\"") // Signet LSP address placeholder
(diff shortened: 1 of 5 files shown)


toneloc:
The PR needs to add an import org.lightningdevkit.ldknode.Network to Constants.kt since the new code references Network.SIGNET and Network.BITCOIN. The LSP pubkey, address, and push registration URL also need to be parameterized per build variant, otherwise debug/signet builds will try to connect to the mainnet LSP and silently fail.

DevAnuragT:
> The PR needs to add an import org.lightningdevkit.ldknode.Network to Constants.kt since the new code references Network.SIGNET and Network.BITCOIN. The LSP pubkey, address, and push registration URL also need to be parameterized per build variant, otherwise debug/signet builds will try to connect to the mainnet LSP and silently fail.

good catch. just pushed an update for this

moved the lsp pubkey, address and push url to build.gradle.kts per variant and updated constants to just pull from buildconfig

i wasn't sure what signet lsp nodes you guys use so i just left them as empty strings in the debug variant for now. let me know if you want me to drop the real ones in there

(btw the Network import was actually already present at the top of Constants.kt on line 5 so it should compile fine now)

PR #31, opened 2026-04-05
fix(ui): correct named parameter in PriceChart drawLine call

DevAnuragT:
PathEffect was passed as a positional argument to drawLine() where
StrokeCap was expected. Added explicit `pathEffect =` named parameter
to fix compiler error.

Discovered while attempting a clean debug build.


toneloc:
Thanks!

PR #32, opened 2026-04-07
fix(iOS): make explorer links network aware

alienx5499:
### Summary
Fix explorer links always pointing to mainnet regardless of configured network.

### Changes
- Added `BitcoinNetwork` enum for safe network handling
- Centralized explorer URL generation in `Constants`
- Replaced hardcoded `mempool.space` URLs in:
  - `HomeView`
  - `SettingsView`
  - `PaymentDetailView`
- Removed force unwrap and used safe optional URL handling

### Implementation
Explorer URLs are now derived from `Constants.defaultNetwork`:

```swift
static func explorerTxURL(txid: String) -> URL? {
    explorerTxURL(txid: txid, network: currentBitcoinNetwork)
}

static func explorerBaseURL(network: BitcoinNetwork) -> String {
    switch network {
    case .mainnet: return "https://mempool.space"
    case .testnet: return "https://mempool.space/testnet"
    case .signet:  return "https://mempool.space/signet"
    }
}
```

### Result
Correct explorer links for mainnet, testnet, and signet.

### DIFF
--- a/ios/StableChannels/StableChannels/Utilities/Constants.swift
+++ b/ios/StableChannels/StableChannels/Utilities/Constants.swift
@@ -1,6 +1,25 @@
 import Foundation
 
 enum Constants {
+    enum BitcoinNetwork: String {
+        case mainnet = "bitcoin"
+        case testnet
+        case signet
+
+        init?(networkString: String) {
+            let normalized = networkString.lowercased()
+            switch normalized {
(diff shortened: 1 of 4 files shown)


alienx5499:
Hi @toneloc, sharing this fix for review.

It resolves explorer links defaulting to mainnet instead of respecting the selected network.

Happy to update based on your feedback.

PR #33, opened 2026-04-08
Fix(iOS): align node runtime network

alienx5499:
### Depends on
#32

### Summary
Start the LDK node using the configured Bitcoin network (`Constants.defaultNetwork`), instead of hardcoding mainnet.

### Changes
- Add `runtimeNetwork` on `AppState` (maps `Constants.BitcoinNetwork` to LDK `Network`)
- Use `runtimeNetwork` for all `nodeService.start` calls:
  - existing wallet startup
  - new wallet creation
  - foreground restart
- Update `SettingsView` to use `appState.runtimeNetwork` instead of `.bitcoin`

### Notes
Includes changes from #32 for stacking. Will rebase after #32 merges.

### Implementation
`AppState`:

```swift
var runtimeNetwork: Network {
    guard let btcNetwork = Constants.BitcoinNetwork(networkString: Constants.defaultNetwork) else {
        assertionFailure("Unsupported defaultNetwork '\(Constants.defaultNetwork)'")
        return .bitcoin
    }
    switch btcNetwork {
    case .mainnet: return .bitcoin
    case .testnet: return .testnet
    case .signet: return .signet
    }
}
```

All `nodeService.start(..., network: ...)` use `runtimeNetwork` or `appState.runtimeNetwork`.

### Result
Node startup uses the configured Bitcoin network (mainnet / testnet / signet).

### DIFF
--- a/ios/StableChannels/StableChannels/AppState.swift
+++ b/ios/StableChannels/StableChannels/AppState.swift
@@ -65,6 +65,21 @@ class AppState {
     private var stabilityTimer: Task<Void, Never>?
     private(set) var chainURL: String = Constants.primaryChainURL
 
+    var runtimeNetwork: Network {
+        guard let btcNetwork = Constants.BitcoinNetwork(networkString: Constants.defaultNetwork) else {
+            assertionFailure("Unsupported defaultNetwork '\(Constants.defaultNetwork)'")
+            return .bitcoin
+        }
+        switch btcNetwork {
+        case .mainnet:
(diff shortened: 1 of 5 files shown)


toneloc:
It looks good but same note as #30: LSP pubkey, address, and push URL are still hardcoded for mainnet, so non-mainnet builds won't connect to an LSP. Those would need to be parameterized per network too. Not sure how to handle it.

Issue #7, opened 2024-04-04
Fix file path


toneloc:
Right now filepath is hardcoded. Could pass in as argument instead. 

CLN has it's own logging conventions for plug-ins. But for right now I think we keep it simple with a json file. 

May want to fix log file so it actually starts as json, instead of a list of JSON objects. In other words, add the beginning and closing brackets.

sagar-h007:
Hi @toneloc, I’d like to take this up.

My plan is to remove the hardcoded file path and make it configurable (probably via a CLI argument with an environment variable fallback), while keeping a sensible default.

I’ll also clean up the logging so it produces valid JSON - initialize the file properly as an array and make sure entries are appended correctly, including handling cases like empty or newly created files.

I’ll keep the changes minimal and stick to the existing style. Let me know if you have a preferred way to handle the configuration.

PR #34, opened 2026-04-08
fix(#7): configure data dir dynamically and format audit log as JSON …

sagar-h007:
### Summary

This PR removes the hardcoded file path and fixes audit logging so it produces valid JSON.

#### Configurable data directory

The data directory is no longer hardcoded.

It now resolves in this order:
	•	--data-dir CLI argument
	•	SC_DATA_DIR environment variable
	•	Default: "data-2/lsp" (kept for backward compatibility)

All usages of the old constant have been updated to use the resolved path. Initialization happens early so everything picks up the correct value.

#### Audit logging fix

Logging previously wrote JSON objects line-by-line, which isn’t valid JSON.

Now it:
	•	Reads existing entries (if the file exists)
	•	Appends the new entry
	•	Writes everything back as a proper JSON array

#### Edge cases
- If the file doesn’t exist -> starts with an empty array
- If the file is empty or invalid -> resets cleanly
- If a write is interrupted -> worst case is losing the last entry, but the file stays valid

Usage
```bash
# Default (no changes needed)
./lsp_backend

# Using environment variable
SC_DATA_DIR=/var/lib/stablechannels/lsp ./lsp_backend

# Using CLI argument
./lsp_backend --data-dir /var/lib/stablechannels/lsp
```
fixs:#7

### DIFF
--- a/src/audit.rs
+++ b/src/audit.rs
@@ -1,4 +1,5 @@
 use serde_json::Value;
+use std::io::Write;
 use std::sync::OnceLock;
 
 static AUDIT_LOG_PATH: OnceLock<String> = OnceLock::new();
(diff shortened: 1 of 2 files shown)


toneloc:
Data dir configuration looks good. Clean fallback chain with CLI/env/default. 

The audit log rewrite is concerning though: reading, deserializing, and rewriting the entire file on every event is O(n) per write and loses the whole log if the process crashes mid-write — the previous append-only approach (JSONL) was more robust, and `jq -s '.'` trivially converts it to a JSON array when needed.

sagar-h007:
@toneloc — Fair point, agreed on both counts.

The rewrite approach is O(n) per event and introduces a corruption window on crash. The original append-only JSONL design is more robust here.

I’ll revert the audit log back to JSONL (one JSON object per line) to preserve append-only durability. Consumers that need a JSON array can use `jq -s '.'`, so this doesn’t need to be handled in the writer.

Updated scope for this PR:
- Keep: configurable data directory (CLI → env → default)
- Revert: audit log back to append-only JSONL

I’ll push an update shortly.

PR #36, opened 2026-04-08
feat(stability): implement riskLevel computation in checkStabilityAction

DevAnuragT:
`riskLevel` existed in `StableChannel` and was already checked as a
circuit breaker inside `checkStabilityAction`. If it exceeded
`MAX_RISK_LEVEL`, no stability payment would be sent. But `riskLevel`
was never actually computed anywhere, so it was always 0 and the
`HIGH_RISK_NO_ACTION` branch was unreachable dead code.

**What this PR does:**

- Added `computeRiskLevel()` to `StabilityService`
- Wired it into `checkStabilityAction` before the action decision
- Risk accumulates when drift exceeds 5% from peg
- Risk decays when the channel is back within the stable band
- The circuit breaker now actually fires during sustained anomalies like a price feed outage or a stuck channel

**Verification:**

Covered by the unit test suite in `test/stability-service-unit-tests`.
The `HIGH_RISK_NO_ACTION`, `STABLE`, and `PAY` tests all exercise this path.

20 tests completed, 0 failed.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/services/StabilityService.kt
+++ b/android/app/src/main/java/com/stablechannels/app/services/StabilityService.kt
@@ -93,6 +93,10 @@ object StabilityService {
         val dollarsFromPar = stableUSDValue - targetUSD
         val percentFromPar = if (targetUSD > 0) abs(dollarsFromPar / targetUSD) * 100.0 else 0.0
 
+        // Update risk level based on current drift before deciding action.
+        // Risk accumulates when far from peg; decays toward 0 when stable.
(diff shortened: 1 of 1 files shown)


DevAnuragT:
While working on this, I noticed that `riskLevel` resets to 0 on 
every app restart since it isn't persisted to the DB. The circuit 
breaker works correctly within a session but loses state on cold starts.

Two options to fix this:
- Add `risk_level` column to the `channels` table (DB_VERSION bump to 3)
- Store it in SharedPreferences since it's per-installation, not per-channel

Happy to follow up with a PR for whichever approach fits better with 
the project direction.

PR #37, opened 2026-04-09
feat(android): add Stable Provider role selection via SharedPreferences

DevAnuragT:
`isStableReceiver` was hardcoded to `true` everywhere in the app,
making the Stable Provider role completely inaccessible. A user who
wants to run as a Provider had no way to set that, even though
`StabilityService` already has correct branching logic for both roles.

**What this PR does:**

- Reads `is_stable_receiver` from SharedPreferences at app init
- Defaults to `true` so existing installations are completely unaffected
- Adds `setStableRole()` in `AppState` to persist and apply the role immediately
- Adds a Role toggle in Settings under the Node card with a description
  of what each role means
- Disables the toggle when a ready channel exists, since switching roles
  mid-channel would invert balance attribution between the two parties

**What stays the same:**

- No DB changes, no migration
- Default behavior is identical to before for all existing users
- The stability logic in `StabilityService` needed no changes since it
  already handles both roles correctly

Also includes the `PriceChart` named-parameter fix to unblock
compilation on this branch, since that fix is pending merge separately.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -56,12 +56,17 @@ class AppState(private val context: Context) : ViewModel() {
     val totalBalanceSats: StateFlow<Long> get() = _totalBalanceSats
 
     init {
-        val prefs = context.getSharedPreferences("balance_cache", Context.MODE_PRIVATE)
+        val prefs = context.getSharedPreferences(Constants.PREFS_NAME_BALANCE_CACHE, Context.MODE_PRIVATE)
         val cachedLightning = prefs.getLong("cached_lightning_sats", 0L)
         val cachedOnchain = prefs.getLong("cached_onchain_sats", 0L)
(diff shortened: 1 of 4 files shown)


toneloc:
Thanks for this; the role toggle logic and the safeguard against mid-channel switching both look well thought out. 

We don't have stable provider UI/UX on the app/client side yet though. So merging this now would expose a setting that doesn't fully work end-to-end.

DevAnuragT:
> Thanks for this; the role toggle logic and the safeguard against mid-channel switching both look well thought out.
> 
> We don't have stable provider UI/UX on the app/client side yet though. So merging this now would expose a setting that doesn't fully work end-to-end.

That makes total sense, I can see how a toggle before the Provider UI is ready could confuse people.

I was digging into the codebase for my Summer of Bitcoin proposal and got this. Since we want to get the end-to-end flow working for both roles, I wanted to explore how to manage the state switch safely.

Should I turn this into a draft for now, or would it be better to close it and merge the logic into the Provider UI PRs later this summer? Just let me know what would be better option.

PR #39, opened 2026-04-11
Remove stale recover/sweep bin targets from Cargo.toml

Antrikshgwal:
This PR removes stale binary target declarations from `Cargo.toml` that pointed to missing source files and caused Cargo to fail while enumerating bin targets.

The manifest declared two bin targets that no longer exist in the repository. As a result, build commands could fail with missing file errors before completing normal compilation.


toneloc:
Thanks. I had these in my gitignore so didn't see the issue.

PR #40, opened 2026-04-16
fix(iOS): avoid nondeterministic channel row selection

alienx5499:
This Pull Request fixes nondeterministic channel row selection in iOS/NSE state reads.

It changes the following:

- Replace `SELECT ... FROM channels LIMIT 1` with deterministic ordering in both iOS paths that read a default channel row:
  - `NotificationService.readChannelState(...)`
  - `DatabaseService.loadChannel(userChannelId: nil)`
- Use `ORDER BY updated_at DESC, channel_id DESC LIMIT 1` to avoid arbitrary row selection when multiple channel rows exist, including timestamp ties.
- Add short inline comments clarifying that this targets deterministic selection (not active-channel semantics).

## Testing
- Built the iOS project and verified no regressions in affected code paths.
- Logic change is limited to SQL ordering; behavior verified via query inspection.

### DIFF
--- a/ios/StableChannels/NotificationService/NotificationService.swift
+++ b/ios/StableChannels/NotificationService/NotificationService.swift
@@ -576,7 +576,14 @@ class NotificationService: UNNotificationServiceExtension {
         defer { sqlite3_close(db) }
 
         var stmt: OpaquePointer?
-        let sql = "SELECT expected_usd, stable_sats, receiver_sats, latest_price, native_sats FROM channels LIMIT 1"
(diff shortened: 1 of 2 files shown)


alienx5499:
Hey @toneloc, would appreciate a quick review on this when you get a chance.

toneloc:
Looks good. Minor note: updateBackingSatsInDB (line 577) still uses MIN(rowid) which could target a different row than the new ORDER BY; worth a follow-up to align them.

alienx5499:
I’ll open a follow-up PR to align that with the deterministic ordering.

PR #41, opened 2026-04-16
fix(tests): update test helpers for ldk-node API changes

Antrikshgwal:
## Context
The project pins a specific `ldk-node` revision in `Cargo.toml`, but `tests/common/mod.rs` wasn't updated to match that revision's API, causing test compilation errors.

## Fixes
- `EsploraSyncConfig` now requires a `timeouts_config` field, added `..Default::default()` to all 3 struct literals in `tests/common/mod.rs`.
- `Builder::build()` now takes a `NodeEntropy` argument, added a `random_entropy()` helper and passed it to all 3 build sites. This mirrors how ldk-node's own integration tests supply per-node entropy.

## How to reproduce
- [ ] `cargo test --no-run` compiles with no errors


toneloc:
Thanks for the contribution.

PR #43, opened 2026-04-19
fix(android): align channel selection for backing sats updates

alienx5499:
This Pull Request follows the iOS deterministic-selection fixes in #40 and #42, and aligns Android push-service channel row selection with the same deterministic policy.

It changes the following:

- Replace nondeterministic fallback read in `StabilityProcessingService.loadChannelStateFromDB(...)`:
  - from `SELECT ... FROM channels LIMIT 1`
  - to `SELECT ... FROM channels ORDER BY updated_at DESC, channel_id DESC LIMIT 1`
- Fix backing sats update target in `StabilityProcessingService.updateBackingSatsInDB(...)`:
  - from `WHERE id = (SELECT MAX(id) FROM channels)`
  - to `WHERE channel_id = (SELECT channel_id FROM channels ORDER BY updated_at DESC, channel_id DESC LIMIT 1)`
- Keep read/write fallback selection policy consistent so both paths target the same deterministic row.

## Testing

- Verified in Android Studio via Gradle sync + Make Project.
- Build succeeded after this change.


toneloc:
Thank you!

PR #44, opened 2026-04-20
feat(android): add UI warning for stale price feeds and fix manual refresh

DevAnuragT:
This PR addresses price feed staleness and improves the manual refresh UX.

**What this does:**
1. **Staleness Warning**: Adds a `isPriceStale` check. If the app has loaded a price but hasn't had a successful API fetch in >60 seconds, a red warning chip appears on the Home Screen indicating the price is stale. This protects users from making decisions on outdated data if they lose connectivity.
2. **First-launch false positives**: Ensured the warning doesn't flash immediately on app start before the first background worker finishes executing.
3. **Wired up Pull-to-Refresh**: Previously, from `HomeScreen`, the pull-to-refresh gesture only refreshed balances but did not actually fire a live network request to the `PriceService`. This now correctly triggers `fetchPrice()`.
4. **Compose UI Animation Glitch**: Added a safe `delay()` baseline to `PullToRefreshBox` to stop the 1.8-alpha spinner from abruptly freezing on the screen when a fetch event resolves instantaneously (like if the background worker happened to just finish a ping).


DevAnuragT:
I am closing this PR, I have added a new and updated PR #78

PR #46, opened 2026-04-20
Fix(fetch_prices): Handle unreachable price feeds and add failure-path tests

### LINKED ISSUE #45: bug(price_feeds): `fetch_prices` aborts on first unreachable feed, UI hangs on "Syncing.."
## Problem

When any single BTC/USD price feed is unreachable (geo-blocked in my case), [`fetch_prices`](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L85) returns `Err` on the first network failure via the `?` operator at [price_feeds.rs:110](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L110), instead of trying the remaining feeds.

```rust
let response = retry(
    Fixed::from_millis(PRICE_FETCH_RETRY_DELAY_MS).take(PRICE_FETCH_MAX_RETRIES),
    || match agent.get(&url).call() {
        // ...
    },
)
.map_err(|e| -> Box<dyn Error> {
    Box::new(std::io::Error::other(e.to_string()))
})?;
let json: Value = response.into_json()?;
```
This propagates the error out of `fetch_prices`, preventing fallback to other configured feeds. [into_json()?](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L112) on the next line has the same short-circuit problem

## Impact

1. Fetch error → fallback to cached price (`0.0` on cold start)
2. `is_syncing` is only cleared when `current_price > 0.0`
3. UI remains stuck on **"Syncing… Fetching latest prices"** indefinitely

Image

## Expected Behavior

Failure of a single feed should not abort the entire fetch. The function should attempt remaining feeds and only return `Err` if all feeds fail.

## Reproduction

1. On a network where at least one configured feed is unreachable (e.g. block `www.bitstamp.net` via `/etc/hosts` or firewall)
2. Run `stable-channels user`
3. App remains stuck on the syncing screen
4. Switching to an unrestricted network unblocks the app

## Proposed Fix

Handle per-feed failures locally instead of propagating via `?`, and continue to the next feed:

- Convert retry errors into a logged `continue 'feeds`
- Only return `Err` if no feeds succeed (already handled by `prices.is_empty()` at `price_feeds.rs:153`)

Antrikshgwal:
Fixes #45.

## Description

[`fetch_prices`](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L85) used `?` on the `retry` result at [price_feeds.rs:110](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L110), so the *first* unreachable feed aborted the whole function. The loop already had a `'feeds:` label and `continue 'feeds` for JSON-shape errors,  but the network-level errors path just never made it there.

Same problem on [`into_json()?`](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L112), one line below: a feed returning unparseable bytes would blow up the whole fetch.

## Fix 

Both are replaced with `match … { Ok => …, Err => { eprintln!(…); continue 'feeds; } }`. The existing `prices.is_empty()` guard at [price_feeds.rs:153](https://github.com/toneloc/stable-channels/blob/656c52bc5cf9685215c46922b701247a001d338f/src/price_feeds.rs#L153) still returns `Err("No valid prices fetched.")` when *nothing* works, so the total-outage contract is unchanged.

## Tests

Two new unit tests in [src/price_feeds.rs](src/price_feeds.rs), both using `std::net::TcpListener` so no new dev-deps:

- `test_one_feed_failure_case` — verifies the one-feed-failure case. Feed list is `[unreachable (127.0.0.1:1), mock server]`; asserts the function returns `Ok` with exactly the mock's price. **Fails on `main`, passes on this branch.**
- `test_fetch_prices_all_unreachable_returns_err` — pins the total-outage contract so a future refactor can't accidentally silently return `Ok(vec![])`.

Existing `#[ignore]`'d `test_fetch_prices_live` kept as-is.

Run:
```
cargo test --lib price_feeds::tests
```

## Follow-ups

- **UX**: If all feeds fail, the UI can remain stuck in a syncing state. A fallback/escape hatch (e.g., “continue anyway”) would help.
- **Price integrity**: In scenarios where 2-3 feeds are unreachable, price manipulation could be a security issue, although I'm not sure if that's urgent or relevant at this stage. Possible improvements:

- require a minimum number of successful feeds (quorum)
- reject outliers beyond a deviation threshold
- enforce freshness checks

Wish to know your thoughts on these.

### DIFF
--- a/src/price_feeds.rs
+++ b/src/price_feeds.rs
@@ -94,7 +94,7 @@ pub fn fetch_prices(
             .replace("{currency_lc}", "usd")
             .replace("{currency}", "USD");
 
-        let response = retry(
+        let response = match retry(
             Fixed::from_millis(PRICE_FETCH_RETRY_DELAY_MS).take(PRICE_FETCH_MAX_RETRIES),
             || match agent.get(&url).call() {
                 Ok(resp) => {
@@ -106,10 +106,21 @@ pub fn fetch_prices(
                 }
                 Err(e) => Err(e.to_string()),
             },
-        )
-        .map_err(|e| -> Box<dyn Error> { Box::new(std::io::Error::other(e.to_string())) })?;
+        ) {
+            Ok(resp) => resp,
+            Err(e) => {
+                eprintln!("Feed {} unreachable: {}", price_feed.name, e);
+                continue 'feeds;
+            }
+        };
 
-        let json: Value = response.into_json()?;
+        let json: Value = match response.into_json() {
+            Ok(v) => v,
+            Err(e) => {
+                eprintln!("Feed {} returned unparseable JSON: {}", price_feed.name, e);
+                continue 'feeds;
+            }
+        };
         let mut data = &json;
 
         for key in &price_feed.json_path {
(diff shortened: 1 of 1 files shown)


toneloc:
Thanks for this contribution @Antrikshgwal

PR #48, opened 2026-04-25
fix(iOS): preserve pending push payment retry state on reconnect failure

### LINKED ISSUE #47: iOS: pending push payment flag is cleared before reconnect outcome
### Summary
On app restart/foreground recovery, iOS can clear `pending_push_payment` before reconnect result is known. If reconnect fails, retry state is lost and pending payment processing can be skipped.

### Repro
1. App receives a push-triggered payment while backgrounded/killed and marks pending payment state.
2. Reopen the app in poor/intermittent connectivity where reconnect may fail.
3. Return to normal connectivity and reopen/foreground again.
4. Observe pending payment recovery may not retry consistently.

### Expected
Keep `pending_push_payment` set until reconnect succeeds; only clear on successful reconnect.

### Impact
- Pending push-payment recovery can be skipped after transient reconnect failures.
- Retry state can be lost even though payment processing has not completed.
- Creates nondeterministic user experience across restart/foreground recovery.

### Scope
| Area | Notes |
| --- | --- |
| Affected | iOS app state recovery path |
| Not in scope | LSP protocol changes, server-side behavior |

### Observed behavior
- Pending payment indicator/state may be cleared even when reconnect does not complete.
- Recovery path can be skipped on later foreground/startup attempts.

### Expected behavior
- Pending payment state should remain until reconnect/payment recovery completes successfully.

alienx5499:
This Pull Request fixes/closes #47

It changes the following:

- Keep `pending_push_payment` set until reconnect outcome is known in `AppState.processPendingPushPayment()`, so reconnect failures do not silently drop retry state.
- Add `AppState.updatePendingPushPaymentFlag(_:reconnectSucceeded:)` to make success/failure flag handling explicit and testable.
- Add `PendingPushPaymentTests.swift` and wire it into `StableChannelsTests` in `project.pbxproj` to cover:
  - legacy drop-before-outcome behavior
  - keep-on-failure behavior
  - clear-on-success behavior

## Testing

- Ran `PendingPushPaymentTests` from Xcode (StableChannelsTests target).
- Without `AppState` helper/change: targeted tests fail (`AppState` missing helper path).
- With `AppState` change restored: targeted tests pass (3/3).

## Scope note

- This change prevents premature retry-state loss by clearing `pending_push_payment` only after successful reconnect initiation.
- It does not guarantee full recovery/payment success; that would require clearing only after confirmed recovery state and is tracked separately.

### DIFF
--- a/ios/StableChannels/StableChannels.xcodeproj/project.pbxproj
+++ b/ios/StableChannels/StableChannels.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		37DA6E9D4F5D4BF36DD86BC8 /* ReceiveView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 445B52C4DEC5CE8D3B0B3DA9 /* ReceiveView.swift */; };
 		3DBEBB9A611388A35C38F1D2 /* MainTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 623F41D841B312F0C9AE01E6 /* MainTabView.swift */; };
 		3E46529F1181F2F866A0A495 /* DatabaseService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3940DC5406ABC63AA090441A /* DatabaseService.swift */; };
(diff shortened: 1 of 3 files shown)


toneloc:
Thanks for this fix Prabal!

PR #67, opened 2026-05-21
fix(android): add gradle wrapper and conditional Firebase plugin

DevAnuragT:
### Summary
  
  The Android project could not build — two critical infrastructure issues were missing from the repo:

  - Gradle wrapper files (gradlew, gradlew.bat, gradle-wrapper.jar) were never committed. Without these, ./gradlew assembleDebug fails immediately.
  - build.gradle.kts unconditionally applied the com.google.gms.google-services plugin, causing the build to fail with File 
  google-services.json is missing.

 Changes

  - **Added Gradle wrapper** — Generated via gradle wrapper --gradle-version 8.11.1. Includes gradlew, gradlew.bat, and gradle-wrapper.jar so anyone can clone and build without installing Gradle separately.
  - **Made Firebase plugin conditional** — build.gradle.kts now only applies the google-services plugin if google-services.json exists. The app builds without it; push notifications simply won't register until the file is
  provided.

  Test plan

  - cd android && ./gradlew assembleDebug succeeds
  - App installs and launches on device without google-services.json


toneloc:
Looks good. Could you please rebase this on current main?

DevAnuragT:
done

PR #71, opened 2026-05-22
Fix: Critical bugs in app

DevAnuragT:
## Summary
  - Fix SQLite crash on splice (UPDATE ... ORDER BY ... LIMIT invalid in SQLite)
  - Fix on-chain deposit detection — poll every 10s until confirmation instead of 60s
  - Fix slider showing on-chain funds — now matches iOS (lightning only)
  - Fix app not refreshing balance on resume — immediate refresh when switching to app
  - Fix Firebase crash when google-services.json missing (try-catch)
  - Add logging for LSP/gateway connect failures
  - Require 3+ price feeds before using price data
  - Add gradle wrapper for reproducible builds

  ## Test plan
  - [x] Send on-chain BTC → "Deposit confirming..." appears immediately on resume
  - [x] After confirmation → "Swap" button appears automatically (no reopen needed)
  - [x] Tap "Swap" → slider updates, no crash
  - [x] Slider only shows lightning balance (on-chain excluded)
  - [x] App starts without crash when google-services.json missing
  - [x] Check logcat for LSP connect failure logs


toneloc:
Thank you for this excellent contribution to the app.

PR #72, opened 2026-05-23
fix(android): backfill hourly prices from Kraken for 1M chart

DevAnuragT:
## Summary
  - Added Kraken OHLC API backfill (matching iOS) to fetch hourly BTC price data for the last 30 days
  - 1M chart now shows actual price data instead of "Collecting price data..."
  - Falls back to daily data when hourly data is insufficient

  ## Changes
  - **PriceService.kt**: Added `fetchKrakenOHLC()` — fetches hourly BTC/USD candles from Kraken API
  - **DatabaseService.kt**: Added `getOldestPriceHistoryTimestamp()` and `backfillHourlyPrices()` — stores Kraken data in price_history table
  - **AppState.kt**: Added `backfillHourlyPrices()` called at startup after `seedHistoricalPrices()`
  - **PriceChart.kt**: Falls back to daily data when hourly data has fewer than 2 points

  ## Test 
  - [x] 1M chart shows correct data

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -134,6 +134,7 @@ class AppState(private val context: Context) : ViewModel() {
 
                 databaseService = DatabaseService(context)
                 launch { databaseService?.seedHistoricalPrices() }
+                launch { backfillHourlyPrices() }
                 tradeService = TradeService(nodeService)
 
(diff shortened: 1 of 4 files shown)


toneloc:
Minor comment from Claude: the ?.let { … null } ?: thirtyDaysAgo chain never short-circuits — since ends up always thirtyDaysAgo, so this re-fetches Kraken on every app start, but INSERT OR IGNORE makes it harmless — merging as-is, worth a one-line follow-up to add an early return.

toneloc:
Thanks for the contribution!

PR #77, opened 2026-05-24
feat(android): add syncing indicator on home screen, enable text selection

DevAnuragT:
## Summary

  Two UX improvements matching iOS behavior:

1.  **Syncing indicator** — Returning users now see the wallet immediately with cached data instead of a full-screen blocking "Syncing wallet..." view. A small "Syncing..." spinner with caption text
  appears below the balance bar until the node finishes syncing. New users without cached data still see the full-screen SyncingView.

2. **Text selection** — Invoice addresses, payment IDs, TXIDs, and addresses are now selectable via long-press. Wrapped detail dialogs and the fund wallet screen in `SelectionContainer`.

  ## Changes

  - `AppState.kt`: Added `isSyncing` StateFlow. When cached channel data exists, skips `Phase.SYNCING` and shows wallet immediately with `isSyncing = true`
  - `HomeScreen.kt`: Added syncing indicator row (small spinner + "Syncing..." caption) below balance bar
  - `FundWalletScreen.kt`: Wrapped address text in `SelectionContainer`
  - `PaymentDetailDialog.kt`: Wrapped dialog content in `SelectionContainer`
  - `TradeDetailDialog.kt`: Wrapped dialog content in `SelectionContainer`

  ## Test plan

  - [x] Returning user: wallet shows immediately + syncing indicator appears then disappears
  - [x] Long-press on address/invoice text: selection handles appear
  - [x] Long-press on payment/trade detail values: text becomes selectable

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -37,6 +37,9 @@ class AppState(private val context: Context) : ViewModel() {
     private val _phase = MutableStateFlow(Phase.LOADING)
     val phase: StateFlow<Phase> = _phase
 
+    private val _isSyncing = MutableStateFlow(false)
+    val isSyncing: StateFlow<Boolean> = _isSyncing
+
     private val _errorMessage = MutableStateFlow("")
     val errorMessage: StateFlow<String> = _errorMessage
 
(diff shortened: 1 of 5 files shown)


toneloc:
Thanks for this Anurag!

PR #78, opened 2026-05-24
fix(android): wire pull-to-refresh to fetch live price, style spinner

DevAnuragT:
## Summary

 Pull-to-refresh on the home screen was broken — it called `refreshBalances()` but never fetched a live price, and the spinner was hardcoded to `isRefreshing = false` so it never showed.

  ## Changes

  - `HomeScreen.kt`: Added `isRefreshing` state that tracks the refresh lifecycle
  - `HomeScreen.kt`: Added `priceService.fetchPrice()` to pull-to-refresh so the price actually updates
  - `HomeScreen.kt`: Added 500ms minimum spinner duration to prevent flash on instant fetches
  - `HomeScreen.kt`: Styled spinner with `PullToRefreshDefaults.Indicator` using primary color

  ## Test plan

  - [x] Pull down on home screen — spinner appears, price updates
  - [x] Spinner uses app theme colors


toneloc:
This required careful testing. Thank you!

PR #79, opened 2026-05-24
fix(android): mark splice payment completed when txid arrives

DevAnuragT:
## Summary

Splice-in payments (sweep to channel) showed as "failed" in payment history even though the sweep completed successfully. The root cause: `setPendingSpliceTxid()` updated the txid but left the status as "pending", which auto-expired to "failed" after 1 hour.

## Changes

- `DatabaseService.kt`: `setPendingSpliceTxid()` now sets `status = 'completed'` alongside the txid update


toneloc:
Good find; I think these stayed in pending before this. Thanks!

PR #80, opened 2026-05-24
fix(android): skip Kraken backfill when price history already covers 30 days

DevAnuragT:
## Summary

  Follow-up to merged chart PR #72  The `?.let { ... null } ?: thirtyDaysAgo` chain never short-circuited — all three code paths evaluated to `thirtyDaysAgo`, re-fetching Kraken on every app start. Added an early return
  when DB already has data covering 30+ days. Refer https://github.com/toneloc/stable-channels/pull/72#issuecomment-4529079127

  ## Changes

  - `AppState.kt`: Early return in `backfillHourlyPrices()` if oldest price timestamp is older than 30 days

  ## Test plan

  - [x] First launch: `CHART_BACKFILL` appears in logcat
  - [x] Second launch: no `CHART_BACKFILL` log (skipped)


toneloc:
Thanks for the fix!

PR #96, opened 2026-05-31
feat(android): add biometric auth, keystore encryption, and clipboard security

DevAnuragT:
## Summary

Adds security infrastructure for biometric authentication, seed phrase encryption, app access preferences, and clipboard security. Closes the biggest security gap — plaintext mnemonic on disk — and adds the foundation for auth gates on sends and seed viewing (wiring comes in a follow-up PR).

## Changes

- `services/BiometricService.kt`: New — BiometricPrompt wrapper with `authenticate()` suspend function, availability checks, lockout fallback to device credential, typed `AuthResult` enum
- `services/KeystoreEncryptionService.kt`: New — AES-256-GCM encrypt/decrypt with Android Keystore hardware-backed key, StrongBox fallback, `migrateFromPlaintext()` for upgrading existing wallets
- `services/AppAccessPreferences.kt`: New — SharedPreferences-backed toggles for App Unlock and Payment Confirmation, `shouldRequireAuth(context, isOnChain)` decision logic
- `util/ClipboardUtils.kt`: New — `copySensitive()` sets `EXTRA_IS_SENSITIVE` on API 33+, schedules clipboard clearing after 60 seconds
- `ui/settings/BackupView.kt`: Added scroll support, clipboard confirmation dialog (styled with rounded corners, warning icon, green Copy button, 60s auto-clear note), blue "Copy Seed Words" button
- `ui/settings/SettingsScreen.kt`: Updated copy button to use clipboard security dialog (legacy screen, kept for compatibility)
- `test/AppAccessPreferencesTest.kt`: Unit tests for auth gate logic — on-chain always requires auth, Lightning only when Payment Confirmation enabled, disabling toggles requires auth

## Test plan

- Settings → Backup → "Backup Seed Words" → seed words appear, page is scrollable
- Scroll down → "Copy Seed Words" button visible (blue outlined)
- Tap "Copy Seed Words" → styled dialog appears with warning icon, "Copy" (green) and "Cancel" buttons
- Tap "Cancel" → dialog dismisses, nothing copied
- Tap "Copy" → seed copied to clipboard, button shows "Copied"
- Paste in another app → seed phrase is there
- Keep app in foreground 60s → paste again → clipboard cleared (empty)
- On Android 13+: clipboard preview in notification shade does NOT show seed text (EXTRA_IS_SENSITIVE)
- App launches normally, no regressions on any existing flows

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/services/AppAccessPreferences.kt
+++ b/android/app/src/main/java/com/stablechannels/app/services/AppAccessPreferences.kt
@@ -1,0 +1,111 @@
+package com.stablechannels.app.services
+
+import android.content.Context
+
+/**
+ * Manages App Access security preferences stored in SharedPreferences.
+ *
+ * Two independent toggles:
+ * - App Unlock: requires biometric auth on launch and resume after 5s in background
+ * - Payment Confirmation: requires biometric auth before Lightning sends
+ *
+ * Auth gate rules:
+ * - On-chain sends ALWAYS require auth (regardless of toggle state)
+ * - Seed phrase viewing ALWAYS requires auth (regardless of toggle state)
+ * - Lightning sends require auth only when Payment Confirmation is enabled
+ * - Disabling either toggle requires auth; enabling does not
+ */
+data class AppAccessPreferences(
+    val appUnlockEnabled: Boolean = false,
+    val paymentConfirmationEnabled: Boolean = false
+)
+
+/**
+ * Utility for reading/writing App Access preferences and determining
+ * when authentication is required.
+ */
+object AppAccessPreferencesManager {
+
+    private const val PREFS_NAME = "app_access_prefs"
+    private const val KEY_APP_UNLOCK = "app_unlock_enabled"
+    private const val KEY_PAYMENT_CONFIRMATION = "payment_confirmation_enabled"
+
+    private fun getPrefs(context: Context) =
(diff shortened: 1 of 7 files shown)


toneloc:
This should not be enabled by default, but the user should be able to opt into it from Settings, similar to how the iOS app exposes App Unlock and Payment Confirmation.

I’m also not sure the actual sensitive flows are wired up yet, such as payment sends or seed reveal; if the first version only gates app open/resume, that’s fine as a starting point. 

The higher priority is making sure the core user flows remain correct and uninterrupted.

We can layer these auth gates onto the specific actions once those flows are very reliable.

DevAnuragT:
Yes there are toggle buttons in settings to enable and disable biometric feature. 
And it will actually be implemented in my next PR which I will post once this is merged. This is like a base, not everything is implemented in this

toneloc:
> Yes there are toggle buttons in settings to enable and disable biometric feature. And it will actually be implemented in my next PR which I will post once this is merged. This is like a base, not everything is implemented in this

Got it. Thanks!

DevAnuragT:
> > Yes there are toggle buttons in settings to enable and disable biometric feature. And it will actually be implemented in my next PR which I will post once this is merged. This is like a base, not everything is implemented in this
> 
> Got it. Thanks!

so can you please merge it so that I can move forward

PR #108, opened 2026-06-03
stable-channels-lsp: TLV trade/sync protocol, splice reconcile, and RegisterPush hardening

Billa05:
Adds interactive stability on the two-daemon split (vanilla LDK Server plus stable-channels-lsp). The wallet drives Buy/Sell and balance changes over a custom Lightning TLV, the daemon verifies and applies them. The core stability math is the shared stable_channels crate, unchanged.

What it adds:
  - TRADE_V1 / SYNC_V1: a node-signed JSON envelope in a custom TLV (type 13377331) over a 1 msat keysend. The daemon reads inbound TRADE_V1 from PaymentReceived custom records, verifies the node signature via LDK Server gRPC, validates the balance, applies the trade, and keysends a signed SYNC_V1 back.
  - Splice reconcile: a splice-out deducts the spliced amount from the stable backing and syncs the wallet.
  - Forwarded-payment reconcile: a forward out of a stable channel deducts the spent amount and syncs.
  - RegisterPush hardening: requires a node-signed REGISTER_PUSH_V1 envelope verified via gRPC inside a 300s window, closing an IDOR where node_id was trusted from the client.
  - Operator GUI: an LDK Server log viewer tab.
  - FCM push: sign the OAuth2 JWT with jsonwebtoken instead of openssl.
  
   Notable differences from server-redo:
 - Splice timing. server-redo deducts pre-confirmation in handle_splice_pending (it reads the new funding output via a bitcoind getrawtransaction RPC) and keeps a post-confirmation reconcile_outgoing backstop. This daemon reconciles post-confirmation only, on the ChannelReady re-fire at splice-lock. It stays event-driven (it consumes the LDK Server SubscribeEvents stream and reacts to the post-splice ChannelReady), it just reconciles at confirmation rather than at broadcast, because the gRPC event stream has no SplicePending event (only ChannelStateChanged PENDING/READY/OPEN_FAILED/CLOSED). It also removes the hardcoded-localhost bitcoind dependency. The only cost is display latency: the wallet shows the last settled expected_usd until the splice locks, then updates in one step. This is arguably safer, since it never deducts for a splice that could still fail. A follow-up could react earlier if upstream adds a SplicePending event over gRPC.
  - RegisterPush is a breaking wire change. The mobile wallet must sign the canonical REGISTER_PUSH_V1 bytes and populate the new signature and timestamp fields, or registration fails closed. Mobile-client signing is a pending follow-up.
  - Forwarded-payment reconcile feeds reconcile_forwarded the pre-forward balance (reconstructed as post_user_sats + total_sats) so native is spent before the overflow eats the peg. server-redo passes the live post-forward balance, which over-deducts when a forward exceeds native. Flagging this for a second look.

### DIFF
--- a/server/lsp-server-gui/src/app.rs
+++ b/server/lsp-server-gui/src/app.rs
@@ -774,6 +774,7 @@ impl LspServerApp {
 						amount_msat,
 						node_id,
 						route_parameters: None,
+						custom_tlvs: vec![],
 					})
 					.await
 					.map_err(|e| e.to_string())
--- a/server/lsp-server-gui/src/app.rs
+++ b/server/lsp-server-gui/src/app.rs
@@ -18,7 +18,7 @@ use sc_rest_client::ldk_server_grpc::api::{
 	VerifySignatureRequest,
 };
 use sc_rest_client::sc_protos::stable::{
-	EditStableChannelRequest, GetPriceRequest, ListStableChannelsRequest,
+	EditStableChannelRequest, GetPriceRequest, ListStableChannelsRequest, LogRequest,
 };
 use sc_rest_client::ldk_server_grpc::types::{
 	bolt11_invoice_description, Bolt11InvoiceDescription, ChannelConfig,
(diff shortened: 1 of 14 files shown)


toneloc:
I like this! I like the RegisterPush hardening .. will need to make some changes before we update LSP, but solid work.

PR #110, opened 2026-06-03
fix(android): show channel close payment as pending until on-chain confirmation

DevAnuragT:
## Problem

When a channel closes, the app immediately marks the channel_close payment as "completed". But the on-chain close transaction may still be unconfirmed in the mempool. The user sees "completed" when funds haven't actually arrived yet.

## Fix

- **User-initiated close:** Payment recorded as "pending", updated to "completed" once `lightningBalanceSats` drops to 0 (meaning the close TX settled on-chain). Home screen shows "Channel closing…" status.
- **Force-close by counterparty:** Payment recorded as "completed" immediately (no reliable way to track confirmation). Home screen shows "Channel closed".

Uses the existing `isChannelClosing` flag lifecycle to determine which case applies.

## Changes

- `android/app/src/main/java/com/stablechannels/app/AppState.kt`


toneloc (approved):
Hope this works.

PR #116, opened 2026-06-04
fix(android): gossip stripping, TLV marker, and isRunning state

DevAnuragT:
### Gossip Stripping
- Before building LDK node in `StabilityProcessingService`, strips `network_graph`, `scorer`, `node_metrics` from SQLite
- Reduces DB from ~10MB to ~30KB, preventing OOM on low-memory devices
- Matches iOS NSE behavior

### TLV Marker Alignment
- Changed stability keysend TLV from `emptyList()` to `listOf(1.toUByte())`
- Sends `[1]` marker byte matching iOS NSE format
- Ensures LSP can validate Android background payments

### Node Status Fix
- `NodeService.isRunning` changed from plain Boolean to `StateFlow<Boolean>`
- NodeView now updates reactively when node starts/stops
- Was showing "Stopped" even when node was running

### DB Schema Fix
- `loadChannelStateFromDB()` referenced non-existent `native_sats` column
- Removed from query, set to 0 in ChannelState

## Files Changed
- `StabilityProcessingService.kt` — gossip stripping, TLV marker, DB fix
- `NodeService.kt` — isRunning StateFlow
- `NodeView.kt` — reactive status display

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/push/StabilityProcessingService.kt
+++ b/android/app/src/main/java/com/stablechannels/app/push/StabilityProcessingService.kt
@@ -78,6 +78,10 @@ class StabilityProcessingService : Service() {
             return
         }
 
+        // Strip gossip from SQLite to avoid OOM in the foreground service.
+        // The NSE doesn't need gossip (it only routes to the LSP, a direct peer).
+        stripGossipFromDB(dataDir)
+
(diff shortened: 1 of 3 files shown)


toneloc:
Do we need the DB gossip stripping on Android?

toneloc:
Thanks. Let's try this out!

PR #118, opened 2026-06-06
feat(iOS): On-chain txid resolution

alienx5499:
## Summary

On-chain deposit resolution via Esplora polling. Close tx resolution via instant LDK API. Shared infrastructure: `ResilientEsploraClient`, `StaggeredTaskLauncher`, `TxidLinkStore`, and `DepositRecorder`. Crash-safe replay.

---

## Problem

- On-chain deposits had no explorer links
- Crash mid-resolution lost pending operations

## Solution

Esplora polling for deposit resolution. Crash-safe replay with shared infrastructure. Extracted testable components: `ResilientEsploraClient`, `StaggeredTaskLauncher`, `TxidLinkStore`, `DepositRecorder`.

---
## Video

https://github.com/user-attachments/assets/28df2ad5-c007-4f69-91e0-546707bbfaec

---

## Architecture Diagram
image

### Architecture Notes

- **ResilientEsploraClient** handles HTTP + jittered backoff + 1MB cap + 15-min budget
- **OnchainTxidResolver** handles receive-address polling for deposits
- **StaggeredTaskLauncher** ensures one task per opId with generation UUIDs
- **Crash-safe**: deposit resolver row written before payment row
- **Replay**: pending ops replayed on app launch

---

1. **On-chain Receive Resolution**
   - Send BTC to receive address
   - Watch `onchain_receive_resolutions` insert and resolve
   - Verify HomeView shows explorer link
   - Check payment row links via `resolution_id`

5. **Unknown Address Flow**
   - Send when no address known
   - Verify bare payment row created (no resolver)
   - HomeView shows pending row without link

6. **Receive Recovery**
   - Force-kill app during receive resolution
   - Relaunch and verify `replayPendingOnchainReceives()` resumes

7. **Expiration Test**
   - Wait 8+ days after resolution
   - Verify `TxidLinkStore` doesn't restore stale txid

### Key Points to Verify

- **Deterministic ordering**: `fetchLatestResolvedOnchainTxid` returns same txid on repeated calls
- **Error handling**: Invalid responses logged but don't crash
- **Memory safety**: No leaks during repeated resolution cycles
- **UI consistency**: All pending rows show appropriate states

## Related
- Closes #105 
- Closes #119

PR #122, opened 2026-06-08
fix(android): close payment ID, txid resolution via Esplora polling

DevAnuragT:
## Problem

1. Channel close payments had the same txid and payment_id as a previous splice out (confusing)
2. Close txid was not available from LDK's `ChannelClosed` event

### Payment ID Fix
- Changed `paymentId` from `fundingTxid` to `channelId` — unique per channel, no collision with splice txids

### Close TX Resolution (inspired by iOS PR #118)
- Created `CloseTxidResolver` — polls Esplora `/tx/{fundingTxid}/outspend/{vout}` to find the close transaction
- When close TX is found, updates the payment with the real txid
- Runs in background with exponential backoff (5s → 10s → 15s → 30s → 60s...)
- Falls back to two chain URLs for reliability

### Database
- Added `updatePaymentTxid()` method to update payment txid after resolution

## How It Works

1. Channel closes → payment recorded with `txid = null`
2. `CloseTxidResolver` starts polling Esplora
3. When funding output is spent → extracts close txid
4. Updates payment with real close txid
5. User sees correct txid in payment history

## Files Changed
- `CloseTxidResolver.kt` — new file, Esplora polling logic
- `DatabaseService.kt` — added `updatePaymentTxid()`
- `AppState.kt` — integrated resolver into `handleChannelClosed()`

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -466,8 +466,9 @@ class AppState(private val context: Context) : ViewModel() {
             // Record in payment history before clearing state
             // If user initiated close, mark pending until on-chain confirms.
             // If force-closed by counterparty, mark completed immediately.
-            val closeTxid = fundingTxid
-            val paymentId = closeTxid ?: channelId
+            // Use channelId as paymentId to avoid collision with splice txids.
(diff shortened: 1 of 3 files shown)


toneloc:
Thanks for aligning this with iOS.

PR #124, opened 2026-06-11
fix(android): UI refresh issues and chain URL fallback

DevAnuragT:
### 1. Balance not refreshed after on-chain send
- After sending on-chain, total balance still showed old amount
- Added `appState.refreshBalances()` after successful send in `OnChainScreen.kt`

### 2. Slider/orders allowed without channel
- Buy/Sell buttons and slider were always enabled, even without a channel
- Added `enabled = hasReadyChannel` to Buy/Sell ActionButtons
- Pass `null` for `onTradeRequest` when no channel (disables slider interaction)

### 3. No feedback on channel close
- After clicking "Close Channel", page was static with no feedback
- Added `setStatus("Closing channel...")` when close is initiated
- Added `refreshBalances()` after successful close
- Added error handling with `setStatus("Close failed: ...")` on failure

### 4. Chain URL fallback not working
- `resolveChainUrl()` was running in parallel, so node started with default URL before fallback was determined
- Changed to resolve chain URL before starting node
- If blockstream.info is unreachable, node now uses mempool.space from the start

## Files Changed
- `HomeScreen.kt` — disabled Buy/Sell buttons and slider when no channel
- `OnChainScreen.kt` — added balance refresh after send
- `ChannelView.kt` — added close feedback and error handling
- `AppState.kt` — resolve chain URL before starting node

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -149,10 +149,8 @@ class AppState(private val context: Context) : ViewModel() {
                 loadChannelFromDB()
                 priceService.startAutoRefresh()
 
-                // Resolve best esplora endpoint in parallel
-                launch {
-                    chainUrl = resolveChainUrl()
-                }
+                // Resolve best esplora endpoint before starting node
+                chainUrl = resolveChainUrl()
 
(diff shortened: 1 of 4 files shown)


toneloc:
I took these and put them in #127 with a bug fix. Thanks.

PR #127, opened 2026-06-12
Fix/UI refresh consolidate

toneloc:


toneloc:
Supersedes #124 with a bug fix @DevAnuragT

PR #130, opened 2026-06-14
fix(android): copy icon visibility and restore dialog UX

DevAnuragT:
### 1. Copy icon invisible on on-chain receive page
- Icon used `MaterialTheme.colorScheme.primary` which was camouflaged with the card's `surfaceVariant` background
- Changed to `MaterialTheme.colorScheme.onSurface` for better visibility

### 2. Restore dialog poor UX
- No loading state — dialog stayed open with no feedback after clicking Restore
- Added `isRestoring` state with progress indicator
- Shows "Restoring wallet..." with spinner while processing
- Disables all buttons during restore
- Prevents dialog dismissal while restoring

## Files Changed
- `FundWalletScreen.kt` — changed icon tint color
- `BackupView.kt` — added loading state to restore dialog


toneloc (commented):
Thanks for this contribution @DevAnuragT

PR #131, opened 2026-06-14
fix(android): deduct from BTC first when sending payments

DevAnuragT:
## Problem

When sending a payment, the app was deducting from the USD (stable) side instead of the BTC (native) side first. This happened because `deductOutgoing` (which takes from BTC first) was never called. Instead, `reconcileOutgoing` ran AFTER the payment and deducted from USD.

## Fix

Added `StabilityService.deductOutgoing()` call BEFORE the payment is sent in `SendScreen.kt`. This ensures:
- Native (BTC) balance is used first
- Only the overflow (if payment > native balance) comes from USD
- `reconcileOutgoing` still runs after payment to reconcile any remaining differences

## Files Changed
- `SendScreen.kt` — added `deductOutgoing` call before payment

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/ui/transfer/SendScreen.kt
+++ b/android/app/src/main/java/com/stablechannels/app/ui/transfer/SendScreen.kt
@@ -3,6 +3,7 @@ package com.stablechannels.app.ui.transfer
 import android.content.ContextWrapper
 import android.widget.Toast
 import androidx.activity.compose.rememberLauncherForActivityResult
+import com.stablechannels.app.services.StabilityService
 import androidx.activity.result.PickVisualMediaRequest
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.foundation.background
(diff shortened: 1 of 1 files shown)


toneloc:
Can you doublecheck this one?

`reconcileOutgoing `already runs after every send and is BTC-first (it only deducts USD for the overflow once native BTC is exhausted), so on success this `deductOutgoing` call is a no-op: the final balances come out identical. 
  
But it mutates `stableChannel.value` the moment the payment is initiated, with no rollback on failure, so a payment that fails will still convert USD→BTC and corrupt the balance. 

If you're seeing USD deducted when it shouldn't be, the real cause is likely `reconcileOutgoing `reading a stale `stableReceiverBTC` before LDK settles, which this wouldn't fix.

DevAnuragT:
> Can you doublecheck this one?
> 
> `reconcileOutgoing `already runs after every send and is BTC-first (it only deducts USD for the overflow once native BTC is exhausted), so on success this `deductOutgoing` call is a no-op: the final balances come out identical.
> 
> But it mutates `stableChannel.value` the moment the payment is initiated, with no rollback on failure, so a payment that fails will still convert USD→BTC and corrupt the balance.
> 
> If you're seeing USD deducted when it shouldn't be, the real cause is likely `reconcileOutgoing `reading a stale `stableReceiverBTC` before LDK settles, which this wouldn't fix.

Got it, so the issue is reconcileOutgoing reading stale stableReceiverBTC before LDK settles. What's the correct way to ensure the balance is fresh? Should we add a delay, or wait for a specific event?

toneloc:
Please monitor this and reopen if you have reproduced the bug and come up with a fix.

DevAnuragT:
Tested on upstream main — payment priority works correctly. When sending, BTC side is deducted first (as expected). The payment failed due to routing issues (RouteNotFound / TemporaryChannelFailure), and the balance correctly reverted after failure.

The real issue is the confirmation screen shows 'Payment sent' optimistically before the payment actually confirms. This is a separate UI bug — will track it separately.

Closing this PR since the payment priority logic is correct.

PR #134, opened 2026-06-15
feat(android): background reliability - WakeLock, gossip stripping, service logging

DevAnuragT:
## Problem

Channels are being force-closed when the app is offline for extended periods. The LSP sends an invalid `channel_reestablish` message when the app reconnects, causing LDK to force-close.

This is directly caused by the app going offline — without FCM push or a background mechanism, the channel goes inactive and the LSP times out.

### 1. WakeLock
- 60-second timeout to prevent CPU sleep during stability payment processing
- Acquired on service start, released in finally block

### 2. Single-Instance Guard
- `AtomicBoolean` prevents duplicate `StabilityProcessingService` invocations
- Second invocation logs "Already running, skipping" and exits

### 3. Gossip Stripping
- Deletes `network_graph`, `scorer`, `node_metrics` before building LDK node
- Saves ~8MB memory, prevents OOM on low-memory devices
- Matches iOS NSE behavior

### 4. Service Logging
- `ServiceLogger.kt` — File-based logger for background service
- Writes timestamped entries to `stability_service.log`
- Capped at 1 MB, evicts oldest entries when full

## How This Prevents Force-Closes

With FCM push working + these reliability improvements:
1. FCM push wakes the app every ~15 minutes
2. Service starts, LDK node syncs, channel stays active
3. No long offline periods → no `channel_reestablish` issues → no force-closes

## Files Changed
- `StabilityProcessingService.kt` — WakeLock, single-instance guard, gossip stripping, logging
- `ServiceLogger.kt` — NEW: File-based logging

## Testing
- FCM push received ✅
- Service starts automatically ✅
- Gossip stripped (8.4MB saved) ✅
- Single-instance guard blocks duplicates ✅
- Service logs written to stability_service.log ✅


toneloc:
I don't think we need this one anymore?

PR #135, opened 2026-06-15
Stable-book sync fix + LSP operator GUI revamp

Billa05:
## Summary
Fixes stable-channel book divergence after an overflow Lightning send, revamps the LSP operator GUI, and aligns the daemon's ldk-node with the workspace root.

## Changes
- **Stable-book sync**:  when an outgoing Lightning send overflows into the USD-pegged portion, the wallet now pushes its new `expected_usd` to the LSP via a signed TLV keysend. The LSP's stability tick gains a balance-truth backstop that self-heals and re-syncs the wallet when a spend went unreconciled, plus an epsilon on the trade balance ceiling and a risk / in-flight-HTLC debounce guard.
- **LSP operator GUI**: dark + amber theme, universal USD/BTC/sats display toggle with unit-aware amount entry, dedicated Settings page, Channels tab gets a two-color liquidity bar, text/status filter, and sortable columns, Payments/Forwarded pagination fixed (no empty-page flicker) with proper loading states; per-tab refresh (removed the global reload). Connection uses native-tls (accepts the daemon's self-signed cert) and the gtk3 file-dialog backend. No daemon API or behavior change.
- **Dependencies**: align the daemon's `ldk-node` rev with the root crate so the workspace builds on current main.

### DIFF
--- a/server/stable-channels-lsp/src/stable_manager.rs
+++ b/server/stable-channels-lsp/src/stable_manager.rs
@@ -70,6 +70,8 @@ pub struct StableChannelManager {
     pub stable_channels: Vec<StableChannel>,
     db: Arc<Database>,
     data_dir: PathBuf,
+    /// Per-channel consecutive low-balance tick count for the balance-truth backstop debounce (ignores transient in-flight HTLCs).
+    spend_debounce: std::collections::HashMap<u128, u8>,
 }
 
 /// Outcome of an `edit_stable_channel` call.
@@ -89,6 +91,7 @@ impl StableChannelManager {
             stable_channels: Vec::new(),
             db,
             data_dir,
+            spend_debounce: std::collections::HashMap::new(),
         }
     }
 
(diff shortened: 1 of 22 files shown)


src/user.rs
toneloc, line 1: So in the case the User is telling the LSP what the expected USD is? Can't the User simply tell the LSP that the USD is higher than it should be? Seems like a cybersecurity issue ... won't work for our model.

server/stable-channels-lsp/src/stable_manager.rs
```diff
@@ -527,6 +537,46 @@ impl StableChannelManager {
             sc.stable_receiver_usd = USD::from_bitcoin(sc.stable_receiver_btc, btc_price);
             sc.latest_price = btc_price;
 
+            // Balance-truth backstop: live balance below backing means a spend went unreconciled (no PaymentForwarded) — deduct + SYNC. Debounced since outbound_capacity excludes in-flight HTLCs.
```
Billa05, line 540: @toneloc This is the source of truth, not the wallet message. Every tick the LSP reads the live channel balance it owns as counterparty and, when it drops below backing, reconciles expected_usd down itself so it's always capped at the user's real sats × price, never what they claim. User-initiated changes still go through a signed TRADE_V1 that the LSP verifies against this same balance (rejected if above holdings).

Because of this backstop we don't actually need the wallet-side push it's just an immediacy optimization (saves one tick), the poll would reconcile on its own. The backstop exists precisely because PaymentForwarded is unreliable and barely fires for these spends, so we can't lean on the event.

So yeah, i will remove the wallet side push

toneloc:
I'm not 100% sure about this one, but let's test it and see how it behaves. 

PR #139, opened 2026-06-19
fix(android): return STABLE when backingSats is zero

DevAnuragT:
## Problem

When `backingSats == 0` and `expectedUSD > 0` (edge case), `checkStabilityAction` falls back to `stableReceiverUSD.amount` (total channel balance) for the drift calculation. This gives an incorrect drift value and could trigger unnecessary stability payments.

## Fix

Return `STABLE` immediately when `backingSats == 0` — no backing means no stable position, so there's no drift to correct.

## Files Changed
- `StabilityService.kt` — added early return when `backingSats == 0`


toneloc:
Thanks for this contribution.

PR #143, opened 2026-06-23
Label stable-channel settlement payments in the operator GUI

### LINKED ISSUE #141: Label stable-channel settlement payments as Top-up / Sync in the LSP operator GUI
Goal: In the operator GUI's Payments tab, label stable-channel settlement keysends as Top-up / Sync instead of the generic "Spontaneous".

Classification: adopt the client's existing marker as a convention:
- Stable-channel keysends carry custom TLV 13377331 (STABLE_CHANNEL_TLV_TYPE), and its value already encodes two signals:
  - 1-byte [0x01] marker → a value (peg) payment → Top-up
  - signed {payload, signature} envelope (TRADE_V1 / SYNC_V1) → a message → Sync
- On the background (push) path, Android and iOS already attach this [0x01] marker to the stability value keysend — so it's effectively already a convention across the clients. (The marker rides the value keysend the wallet sends; the FCM/APNs push is only a wake and carries no TLV.) If we keep it as the convention, the LSP can simply read the marker to classify settlements — no new envelope or protocol. We also bring the desktop wallet onto the same convention:
  - Android: byteArrayOf(1) https://github.com/toneloc/stable-channels/blob/832a23f2ef5be74b7ae9a77cb5101fab5f249569/android/app/src/main/java/com/stablechannels/app/push/StabilityProcessingService.kt#L273
  - iOS: Data([1]) https://github.com/toneloc/stable-channels/blob/832a23f2ef5be74b7ae9a77cb5101fab5f249569/ios/StableChannels/NotificationService/NotificationService.swift#L476

Plan:
- The TLV isn't in ListPayments (the Spontaneous kind is hash + preimage only), only the daemon sees it at event time, so the daemon records each settlement's (payment_id, kind):
  - inbound: split on the TLV value envelope → sync (+ apply trade), marker → top_up
  - outbound: at the LSP's own top-up (run_tick) and sync (send_sync_message) sites
- New settlement_payments table → ListSettlementPayments REST endpoint (HMAC-auth) → REST client → GUI fetches the payment_id → kind map alongside payments and relabels matching rows Top-up / Sync.

Caveats:
- Foreground gap (both mobile clients): 
Android https://github.com/toneloc/stable-channels/blob/832a23f2ef5be74b7ae9a77cb5101fab5f249569/android/app/src/main/java/com/stablechannels/app/AppState.kt#L571 and iOS https://github.com/toneloc/stable-channels/blob/832a23f2ef5be74b7ae9a77cb5101fab5f249569/ios/StableChannels/StableChannels/App/AppState.swift#L1316 send the foreground stability payment as a plain keysend with no TLV; only the background (push / NSE) path attaches the marker. So settlements sent while the app is actively open aren't tagged until each gets a one-line fix (sendKeysend → sendKeysendWithTLV + marker). Background-sent settlements (the common case) are tagged.

Billa05:
Stable-channel settlement keysends now show as **Settlement** (value/peg) or **Sync**
(state message) in the Payments tab, instead of the generic "Spontaneous".

The discriminator is custom TLV `13377331`, visible only at the event layer, not in
`ListPayments`. So the daemon classifies by the marker it already carries (no new envelope):
`[0x01]` → Settlement, signed envelope (TRADE_V1/SYNC_V1) → Sync.

- Daemon records `(payment_id, kind)` → new `settlement_payments` table (`src/db.rs`)
- `ListSettlementPayments` REST endpoint → `sc-rest-client` → GUI relabels matching rows
- Desktop wallet brought onto the same marker (`src/stable.rs`)

**Caveats:** mobile *foreground* path doesn't mark yet (background does), one-line client follow-up.

**Naming** (Settlement/Sync) per #141, open to final wording.

Closes #141.

### DIFF
--- a/server/lsp-server-gui/src/app.rs
+++ b/server/lsp-server-gui/src/app.rs
@@ -17,7 +17,7 @@ use sc_rest_client::ldk_server_grpc::api::{
 	SpliceInRequest, SpliceOutRequest, SpontaneousSendRequest, UpdateChannelConfigRequest,
 	VerifySignatureRequest,
 };
-use sc_rest_client::sc_protos::stable::{GetPriceRequest, ListStableChannelsRequest, LogRequest};
+use sc_rest_client::sc_protos::stable::{GetPriceRequest, ListSettlementPaymentsRequest, ListStableChannelsRequest, LogRequest};
 use sc_rest_client::ldk_server_grpc::types::{
 	bolt11_invoice_description, Bolt11InvoiceDescription, ChannelConfig,
 };
@@ -247,6 +247,7 @@ impl LspServerApp {
 					.map_err(|e| e.to_string())
 			}));
 		}
+		self.fetch_settlement_payments();
 	}
 
 	pub fn fetch_peers(&mut self) {
@@ -753,6 +754,21 @@ impl LspServerApp {
 		}
 	}
 
+	pub fn fetch_settlement_payments(&mut self) {
+		if self.state.tasks.list_settlement_payments.is_some() {
+			return;
+		}
+		if let Some(client) = &self.state.client {
+			let client = client.clone();
+			self.state.tasks.list_settlement_payments = Some(self.spawn_task(async move {
+				client
+					.list_settlement_payments(ListSettlementPaymentsRequest {})
+					.await
+					.map_err(|e| e.to_string())
+			}));
+		}
+	}
+
 	pub fn disconnect_peer(&mut self, node_pubkey: String) {
 		if self.state.tasks.disconnect_peer.is_some() {
 			return;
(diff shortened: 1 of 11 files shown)


Billa05:
For this PR let's scope to 2 labels: they're the only two things the daemon can actually tell apart rn (by the TLV on the keysend):
- Stability: the value keysend either side sends to move sats and keep the channel at its expected-USD peg.
 - Sync: the signed state messages: the wallet's TRADE (sets a new expected-USD) and the LSP's SYNC (tells the wallet the peg changed after a spend).
 
"User" and "Fee" would be a different, broader categorization of all payments, the daemon can't detect those today, so I'd do that as a separate pass later. Sound good @toneloc?

PR #145, opened 2026-06-28
reject stale signed TRADE_V1 messages (replay protection)

Billa05:
Bounds the validity window of signed TRADE_V1 trades so a captured/leaked envelope can't be re-applied indefinitely.

A signed TRADE_V1 {payload, signature} is currently valid forever, so anyone holding a copy (LSP logs, wallet storage, backups, a compromised endpoint) can mint a fresh keysend carrying the old blob and re-apply a stale trade, overriding
the channel's current expected_usd. This is defense-in-depth, the Lightning onion already blocks passive wire capture; the gap is the unbounded lifetime of the signed command, not a wire MITM.

- Add `ts: u64` to the signed TRADE_V1 payload.
- After signature verification, reject if |now - ts| > 300s (audit: TRADE_STALE).
- Gated on `ts != 0`, so un-upgraded wallets (ts absent) still apply as before, backward compatible.

Server half of a two-part change: full effect lands once the wallet signs a real `ts`; flip the gate to reject `ts == 0` after wallet rollout.

Also removes the unused `new_without_cert` REST-client constructor (dead code).

### DIFF
--- a/server/sc-rest-client/src/client.rs
+++ b/server/sc-rest-client/src/client.rs
@@ -112,22 +112,6 @@ impl LspRestClient {
 		Ok(Self { base_url, client, api_key, use_relative_urls: true })
 	}
 
-	/// Constructs a [`LspRestClient`] without requiring a TLS certificate.
-	///
-	/// This is useful for WASM targets where the browser handles TLS, or for
-	/// development environments where certificate verification is handled differently.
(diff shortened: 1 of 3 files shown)


toneloc:
Thank you. We will need to take closer look when merge the ldk-server "server-redo" branch into this finally.

PR #147, opened 2026-06-30
feat(android): classify incoming rebalancing payments using custom TLVs

DevAnuragT:
Classifies incoming Lightning payments containing the custom TLV marker (type 13377331) as stability payments instead of lightning. This ensures alignment with iOS and recent LSP-side updates that tag rebalancing keysends with a custom TLV marker byte, allowing correct classification as Settlement in the payment history.


toneloc:
We were working on a lot of the same files on this one. #146 included many fixes so I will close this for now.

PR #155, opened 2026-07-04
Robust, searchable audit log

Billa05:
## Summary

The LSP already writes an audit log, but it was lossy (dropped channel-close reasons, lifecycle boundaries, and DB/gRPC failure branches), flood-prone (a line per price fetch, and per-tick stability heartbeats), and hard to query (tail-N only, single-id). This PR makes it a record you can actually use to reconstruct a channel's whole lifecycle, trims the noise, and adds operator UI to read it.

## 1. What the log now captures

Events are keyed by **both** the hex `channel_id` and the decimal `user_channel_id` where a channel is involved, so filtering by either returns a channel's complete history.

**Channel lifecycle**
- `CHANNEL_PENDING`, `CHANNEL_OPEN_FAILED`, previously silent
- `CHANNEL_READY_TRACKED`, `CHANNEL_READY_SPLICE`
- `CHANNEL_CLOSED`, now with full close reason (kind, initiator, peer message, counterparty, funding txo); force-closes flagged
- `CHANNEL_STATE_UNKNOWN`, catch-all for unmapped/future states
- `CHANNEL_MARKED_CLOSED_AT_STARTUP`

**Payments**
- `PAYMENT_RECEIVED`, `PAYMENT_CLAIMABLE`
- `PAYMENT_SETTLED`, `PAYMENT_FAILED`, correlated back to a channel via a send-time settlement→channel map
- `PAYMENT_FORWARDED` (both legs keyed), `PAYMENT_FORWARDED_BACKFILL`

**Stability (peg maintenance)**
- `STABILITY_PAYMENT_SENT` / `_FAILED`, `STABILITY_PUSH_QUEUED`, `STABILITY_CHECK_ONLY`, `STABILITY_SKIP_HIGH_RISK`, `STABILITY_COOLDOWN`
- `BACKSTOP_STABLE_DEDUCTED`, `STABLE_SPEND_DEDUCTED`, `SPLICE_OUT_STABLE_DEDUCTED`, `STABLE_EDITED`

**Trades** (full validation ladder)
- `TRADE_PARSED_PAYLOAD_OK`, `TRADE_PARSE_PAYLOAD_FAILED`, `TRADE_PARSE_SIGNED_FAILED`, `TRADE_SIGNATURE_VALID` / `_INVALID`, `TRADE_APPLIED`, `TRADE_STALE`, `TRADE_EXCEEDS_BALANCE`, `TRADE_INVALID_AMOUNT`, `TRADE_CHANNEL_NOT_FOUND`, `TRADE_CHANNEL_UID_UNPARSEABLE`, `TRADE_STABLE_ENTRY_NOT_FOUND`, `TRADE_UNHANDLED_TYPE`

**Sync / push**
- `SYNC_MESSAGE_SENT` / `_FAILED`, `MESSAGE_RECEIVED`, `REGISTER_PUSH_OK`, `REGISTER_PUSH_SIGNATURE_INVALID`

**Peers & on-chain sweeps** (periodic reconcile-from-truth poll)
- `PEER_CONNECTED`, `PEER_DISCONNECTED`, `SWEEP_PROGRESS`

**Infrastructure failures** (previously trace-only)
- `DB_WRITE_FAILED`, `DB_READ_FAILED`, `LDK_CALL_FAILED`

Completeness after reconnects doesn't rely on a durable event queue. On each event-stream (re)connect, the daemon reconciles against LDK Server's durable stores and backfills any forwards it hadn't seen (deduped by a persistent fingerprint; audit-only, no double peg-math).

## 2. Making it readable: removing noise

The log was dominated by two high-frequency, low-value sources:
- **`PRICE_FETCH`**: one line per price poll. Removed entirely.
- **Repetitive stability lines**: the daemon logged its stability evaluation every tick even when nothing changed. Now **edge-triggered**: it logs on an outcome change or a material value move past the existing thresholds, not on every tick. Money-moving events always log.

Principle: an audit log should record **events and transitions, not a per-tick heartbeat**. Every transition is still captured, so durations stay reconstructable. This changes emission cadence only, never any stability decision.

## 3. Channel History (readability + search)

- The audit-log endpoint gained a **server-side substring filter** and a **full-history mode** (complete matching history, oldest-first, capped).
- Operator GUI adds **Audit | Channel History | LDK server** log tabs: a plain audit tail, and a **Channel History** tab that takes a channel id and shows that channel's entire timeline oldest-first. Force-closes render with a marker.

## 4. Reliability fixes

- **Audit-writer atomicity:** writing a JSON value with `writeln!` serialized token-by-token, so concurrent writers could interleave into corrupt lines. Each line is now built and written with a single atomic append. (Shared crate, fixes the desktop wallet too.)
- **Cold-start reconcile:** the in-memory stable-channel list could stay empty after a restart if the price cache was still cold when the startup/first-reconnect reconciles ran. Startup now waits for the first price before reconciling, with a per-tick backstop that rebuilds an empty list (and leaves a populated one untouched).

## Compatibility

- New audit-log request fields (`filter`, `full`) are additive and default to the prior behavior when omitted.
- The settlement DB migration is idempotent; the original settlement API signature is preserved.
- No changes to LDK Server's proto/gRPC, only the LSP's own REST contract and its handling of already-delivered events.

## Still missing, needs an LDK Server (upstream) PR

These blind spots require fields/events LDK Server doesn't expose yet, so they're deliberately out of scope here:

- **Per-hop payment failure reasons:** `PaymentFailed` / payment details carry no failure-reason field.
- **Splice lifecycle events:** splice pending/failed are swallowed by LDK Server's event catch-all (currently observability-only, since a splice re-ready already triggers reconcile).

## Future work

- **SQLite-backed audit store:** would let the stability throttle be replaced with write-everything + collapse-on-read, and make queries first-class (indexed by channel/event/time) instead of scanning a file.

### DIFF
--- a/server/lsp-server-gui/Cargo.toml
+++ b/server/lsp-server-gui/Cargo.toml
@@ -16,6 +16,7 @@ sc-protos = { path = "../sc-protos" }
 futures-util = "0.3"
 toml = "0.8"
 serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
 hex = { package = "hex-conservative", version = "0.2", default-features = false, features = ["std"] }
 bytes = "1.4.0"
 
--- a/server/lsp-server-gui/src/app.rs
+++ b/server/lsp-server-gui/src/app.rs
@@ -286,7 +286,34 @@ impl LspServerApp {
 		if let Some(client) = &self.state.client {
 			let client = client.clone();
 			self.state.tasks.ldk_log = Some(self.spawn_task(async move {
-				client.ldk_log(LogRequest { max_lines }).await.map_err(|e| e.to_string())
+				client.ldk_log(LogRequest { max_lines, filter: String::new(), full: false }).await.map_err(|e| e.to_string())
+			}));
+		}
+	}
+
+	pub fn fetch_audit_log(&mut self) {
+		if self.state.tasks.audit_log.is_some() {
+			return;
+		}
+		let max_lines = self.state.forms.audit_log.max_lines.parse::<u32>().unwrap_or(200);
+		if let Some(client) = &self.state.client {
+			let client = client.clone();
+			self.state.tasks.audit_log = Some(self.spawn_task(async move {
+				client.audit_log(LogRequest { max_lines, filter: String::new(), full: false }).await.map_err(|e| e.to_string())
+			}));
+		}
+	}
+
+	/// Complete history for the filtered id, oldest-first (server `full` mode, ignores max_lines).
+	pub fn fetch_channel_history(&mut self) {
+		if self.state.tasks.channel_history.is_some() {
+			return;
+		}
+		let filter = self.state.forms.channel_history.filter.clone();
+		if let Some(client) = &self.state.client {
+			let client = client.clone();
+			self.state.tasks.channel_history = Some(self.spawn_task(async move {
+				client.audit_log(LogRequest { max_lines: 0, filter, full: true }).await.map_err(|e| e.to_string())
 			}));
 		}
 	}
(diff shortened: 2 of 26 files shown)


toneloc:
Nice progress. Thank you for this contribution.

PR #169, opened 2026-07-09
feat(iOS): NSE for Stability

### LINKED ISSUE #120: iOS: NSE for Stability
## Problem 

- NSE has limited resources (~30s CPU, ~30–50MB RAM)
- Full LDK gossip DB (~10MB) causes OOM in NSE
- NSE and main app write to same Lightning DB → risk of state mismatch → force closes
- Need DB-write coordination between app and NSE

alienx5499:
## Summary

Massive architectural refactor of the iOS Notification Service Extension (NSE). Breaks down the monolithic 1400-line `NotificationService.swift` into focused, testable, single-responsibility components with strict protocol boundaries.

---

## Problem

The original `NotificationService.swift` contained all business logic, database operations, LDK node management, price fetching, and UI content formatting in a single massive file. This led to:
- A giant, untestable switch statement for handling payment directions.
- Direct tightly-coupled infrastructure dependencies.
- Difficult-to-track exit paths causing potential lock leaks.
- Duplicated parsing logic across different handlers.
- High risk of regressions when making minor modifications.

## Solution

1. **Protocol-Oriented Design**: Extracted all external dependencies behind protocols (`PaymentDatabase`, `PriceFetcher`, `NodeStarter`, `FileLogger`, `NotificationContentMutator`).
2. **Strategy Pattern for Payments**: Created a `PaymentHandler` protocol and a factory that delegates payment processing to dedicated classes (`IncomingPaymentHandler`, `UserToLSPHandler`, `LSPToUserHandler`).
3. **Domain Models**: Extracted constants and types (`PaymentDirection`, `StableControlParser`) into their own files.
4. **Database Safety & Locking**: Built `SQLitePaymentDatabase` with atomic cross-process operations. Implemented a robust database locking mechanism using a shared lock file to prevent race conditions, unifying locking across the `NotificationService` and `DatabaseService`.
5. **Enhanced UX**: Added default notification sounds for incoming and outgoing payment alerts, and refreshed payment status on app launch.
6. **Orchestrator Pattern**: Reduced `NotificationService.swift` to a slim orchestrator that purely initializes dependencies and manages the high-level lifecycle.

---

### Modified

- `ios/StableChannels/NotificationService/NotificationService.swift`
  - Removed 1200+ lines of code; refactored to use Dependency Injection.
- `ios/StableChannels/StableChannels.xcodeproj/project.pbxproj`
  - Registered all new NSE service files into the Xcode target.

---
### Related
- Closes #120

### DIFF
--- a/ios/StableChannels/NotificationService/NotificationService.swift
+++ b/ios/StableChannels/NotificationService/NotificationService.swift
@@ -68,6 +68,9 @@ class NotificationService: UNNotificationServiceExtension {
             return
         }
 
+        // Group payment notifications under a single thread to coalesce them
+        content.threadIdentifier = "stable-channels-payment"
+
         // Parse direction from push payload
         let userInfo = request.content.userInfo
         let direction: String
@@ -83,6 +86,20 @@ class NotificationService: UNNotificationServiceExtension {
         } else {
             direction = "lsp_to_user"
         }
+
+        // Tag with a direction-specific type for logic checks in foreground app
+        let notificationType: String
+        switch direction {
+        case "user_to_lsp":
+            notificationType = "payment_sent"
+        case "incoming_payment":
+            notificationType = "payment_received"
+        default: // "lsp_to_user"
+            notificationType = "payment_received"
+        }
+        var updatedUserInfo = content.userInfo
+        updatedUserInfo["notification_type"] = notificationType
+        content.userInfo = updatedUserInfo
 
         nseLog("didReceive: direction=\(direction) userInfo=\(userInfo)")
 
--- a/ios/StableChannels/NotificationService/NotificationService.swift
+++ b/ios/StableChannels/NotificationService/NotificationService.swift
@@ -703,6 +703,7 @@ class NotificationService: UNNotificationServiceExtension {
                     dollarsAbs
                 )
                 UserDefaults(suiteName: Self.appGroup)?.set(false, forKey: "pending_push_payment")
+                content.sound = .default
             case .failed, .missingChannelRow:
                 nseLog("Payment sent but DB persistence failed — durable guard will block resends")
                 content.title = "Payment Sent"
--- a/ios/StableChannels/NotificationService/NotificationService.swift
+++ b/ios/StableChannels/NotificationService/NotificationService.swift
@@ -132,7 +132,8 @@ class NotificationService: UNNotificationServiceExtension {
         }
 
         // Acquire DB lock before any operations to prevent race with main app
-        let lockFile = container.appendingPathComponent("nse.db.lock")
+        // Uses the same App Group lock file as DatabaseService (stablechannels.db.lock).
+        let lockFile = container.appendingPathComponent("stablechannels.db.lock")
         var lockFD: Int32 = -1
         lockFile.withUnsafeFileSystemRepresentation { path in
             if let path {
--- a/ios/StableChannels/NotificationService/NotificationService.swift
+++ b/ios/StableChannels/NotificationService/NotificationService.swift
@@ -3,56 +3,24 @@ import LDKNode
 import SQLite3
 import Darwin
 
-/// Notification Service Extension — starts a lightweight LDK node to handle
-/// stability payments while the main app is killed.
-///
-/// Push payload includes `stability.direction`:
-/// - `lsp_to_user`: LSP owes user sats (price dropped). Start node, wait for incoming payment.
-/// - `user_to_lsp`: User owes LSP sats (price rose). Start node, calculate amount, send keysend.
+/// Notification Service Extension — handles stability payments while main app is killed.
+/// Uses dependency injection for testability and SOLID compliance.
 class NotificationService: UNNotificationServiceExtension {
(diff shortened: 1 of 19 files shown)


toneloc:
In review

toneloc:
Thanks for this contribution; it's a big overhaul so please check these comments. 

The direction is good, but do not think we can merge this as-is. This changes cross-process locking, event-loop behavior, pending-payment state, and foreground notification behavior. Two of those are blocker-level for the Notification Service Extension's core job.

## Blockers

1. The new `flock` scheme can block the NSE in the normal background case.

`DatabaseService.init` now takes an exclusive `flock` on `stablechannels.db.lock` and holds it for the main app process lifetime. The NSE then tries to take the same blocking lock before processing a push. On iOS, the common state is: app backgrounded, node stopped, process still suspended/resident, and the DB service still alive. The heartbeat gets stale, so `isMainAppActive()` passes, but the NSE blocks on `flock` until the extension times out. That defeats the background-push path the NSE exists for.

Relevant spots:

- `ios/StableChannels/StableChannels/Services/DatabaseService.swift` around `init` / `deinit`
- `ios/StableChannels/NotificationService/NotificationService.swift` around `acquireDBLock`

Also, `cleanup()` releases the lock before `node.stop()`. If the lock is meant to guard LDK data-dir/node ownership, that ordering defeats it.

Please replace this with a narrower node-owner guard, or remove the broad `flock` layer and continue relying on the existing SQLite `BEGIN IMMEDIATE` transactions for payment DB atomicity. Whatever guard remains should be nonblocking or short-timeout, tied to actual node ownership, and released only after the NSE node has fully stopped.

2. The new event loops can hot-spin on an unacknowledged event.

`IncomingPaymentHandler` and `LSPToUserHandler` now use an inner `while let event = node.nextEvent()` loop, with the 22s timeout and sleep only in the outer loop. But when DB persistence fails, the code intentionally does not call `node.eventHandled()` so LDK can redeliver after a durable write succeeds. That safety behavior is correct, but with the new inner loop the same unhandled event can be returned repeatedly without hitting the outer timeout or the `0.5s` sleep. Result: tight retry loop, no completion, and `node.stop()` may not be reached.

Please restore the old semantics: check the timeout, process at most one event, sleep, and exit cleanly at the deadline. If you want to drain multiple events, any intentionally-unacked persistence failure needs to break back out to the timeout/sleep path.

## Other issues to fix before merge

3. Idle incoming wakes now create false pending notifications.

The old no-payment `incoming_payment` wake delivered empty/silent content and did not mark `pending_push_payment`. The new code turns `!received` into `"Payment Pending" / "Open app to receive your payment"` and returns `shouldPersist = true`. That will create phantom pending alerts/state for wake pushes where no payment arrived.

Please restore the silent/no-persist idle-wake behavior unless there was an actual persistence failure or a control message that must be deferred to foreground.

4. `pending_push_payment` is now overwritten too broadly.

The orchestrator writes `pending_push_payment = shouldPersist` on every handler completion. That means no-op paths like cooldown, within-threshold, or "position stable" can clear an existing `true` flag from a previous failed payment. Please only clear that flag when this run actually resolved the pending condition; otherwise preserve an existing `true`.

5. `applySyncMessage` lost the live-price fallback.

The old code fetched a fresh price if `latest_price <= 0` before recomputing backing. The new `SQLitePaymentDatabase.applySyncMessage` preserves current backing in that case while still updating `expected_usd`, which can leave stale `stable_sats` / `native_sats` feeding future drift checks. Please restore the price fallback or explicitly defer the sync to foreground when price is unavailable.

6. `incoming_payment` stops after the first received batch.

The old incoming path kept polling the full NSE window to catch additional payments. The new handler breaks once `received` is true. Please keep polling until timeout unless a stable-control path explicitly stops processing.

7. Please split unrelated main-app behavior changes out of this PR.

The foreground payment-banner suppression, notification content/sound changes, `HistoryView` reload behavior, and main-app `DatabaseService` locking redesign make this much harder to review as an NSE refactor. Those should be separate PRs unless they are strictly required for the NSE extraction.

## What looks good and should be preserved

- Handler/protocol decomposition is a good direction.
- `pending_stability_send` claim/send/record/clear flow is broadly preserved.
- Stability thresholds, TLV type, keysend construction, msat/sat conversions, and backing deltas look faithful.
- `SQLITE_TRANSIENT` is a real correctness improvement.
- The DispatchSource heartbeat is better than the old timer.
- New NSE files appear to be in the right target, and the simulator build passes.

## Before merge, I would want device coverage for:

- push while the app process is backgrounded/suspended but still resident
- push while the app is killed
- cold app launch while the NSE is processing
- forced DB persist failure / missing channel row during incoming payment
- idle `incoming_payment` wake with no payment
- multiple incoming payments in one NSE window
- `SYNC_V1` when `latest_price` is zero/missing
- `user_to_lsp` cooldown while `pending_push_payment` is already true

Bottom line: I like the refactor shape, but the current runtime behavior is riskier than the monolith. Fix the locking and event-loop blockers plus the pending-state regressions, and I think this can become a solid improvement.

toneloc:
Thanks for this refactor.

PR #173, opened 2026-07-13
iOS force close issues

toneloc:
We are facing force closes on iOS when processing payments through notifications. This will try to fix it.


toneloc:
Pushing this to main.

PR #174, opened 2026-07-14
Force close fixes

toneloc:
Multiple changes to add locks to fix force close issues


toneloc (commented):
Tested on iOS and Android

PR #175, opened 2026-07-14
Bound HTTP client timeouts (wallet, FCM push, REST client)

Billa05:
`ureq::Agent::new()` and reqwest's default client build with no timeout, so a hung/geo-blocked endpoint blocks forever. This is what froze the wallet's stability tick when a price feed stopped responding (geo-blocked Bitstamp), the fetch never returned, so the loop never ticked.

Adds bounded connect + overall timeouts to every live HTTP client:

- Wallet (price_feeds.rs, user.rs): new shared bounded_agent() (5s connect / 15s) for the esplora + Kraken chart/fee fetches; the per-tick price agents keep their tighter 5s/10s.
- Daemon FCM push (fcm.rs): new http_client() helper (10s / 20s) for both the push send and the OAuth token request.
- REST client (sc-rest-client): 10s connect / 60s on the client builder.

### DIFF
--- a/server/sc-rest-client/src/client.rs
+++ b/server/sc-rest-client/src/client.rs
@@ -91,6 +91,8 @@ impl LspRestClient {
 
 		let client = Client::builder()
 			.add_root_certificate(cert)
+			.connect_timeout(std::time::Duration::from_secs(10))
+			.timeout(std::time::Duration::from_secs(60))
 			.build()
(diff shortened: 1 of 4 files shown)


Billa05:
@toneloc could you please review this

toneloc:
Good work. Thanks for this improvement.

PR #176, opened 2026-07-14
Gate GUI deps behind a gui feature

Billa05:
The headless LSP daemon pulled in the entire desktop GUI stack `(eframe/egui/winit/glow/image/qrcode)` because the root stable-channels crate's GUI deps were non-optional.

Fix (manifest-only):
- Mark those 5 deps optional, gated behind a new gui feature (kept in default, so the wallet is unaffected).
- Daemon depends on stable-channels with default-features = false.

The lib is GUI-free, only the wallet bin uses egui, so nothing needed `#[cfg]` gating.

Verified with from-scratch builds:
- cargo build -p stable-channels-lsp -> 0 GUI crates (108 fewer crates total).
- cargo build --bin stable-channels -> still builds with egui.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,8 @@ version = "0.9.0"
 edition = "2021"
 
 [features]
-default = []
+default = ["gui"]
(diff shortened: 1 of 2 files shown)


toneloc:

PR #180, opened 2026-07-15
feat(android): UI Polish, Explorer Links, and Status Capsule Updates

DevAnuragT:
This PR introduces:
1. Reduced spacing across HomeScreen components (BalanceBar, PriceChart, etc.) for a more compact and modern aesthetic.
2. A completely robust draggable tooltip via a Compose Popup window to prevent clipping at the screen edges.
3. Smart routing for the Status Capsule to open either Trade or Payment Details depending on message context (including channel closes).
4. A new OnchainTxidResolver that automatically adds clickable Mempool.space explorer links to pending on-chain deposit cards.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -1511,8 +1511,7 @@ class AppState(private val context: Context) : ViewModel() {
         // Clear closing flag once lightning balance fully resolves, or if a new channel is opened
         // Don't clear pendingClosePaymentId here — let detectOnchainDeposit()
         // handle it when the on-chain funds arrive
-        val isOpeningChannel = _stableChannel.value.userChannelId.isNotEmpty() && !hasReady
-        if (isChannelClosing && (lightning == 0L || hasReady || isOpeningChannel)) {
+        if (isChannelClosing && lightning == 0L) {
             isChannelClosing = false
         }
 
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -76,9 +76,15 @@ class AppState(private val context: Context) : ViewModel() {
 
     private val _totalBalanceSats: MutableStateFlow<Long>
     val totalBalanceSats: StateFlow<Long> get() = _totalBalanceSats
-
     private val _hasReadyChannel = MutableStateFlow(false)
     val hasReadyChannel: StateFlow<Boolean> get() = _hasReadyChannel
+
+    private val _onchainReceiveAddress = MutableStateFlow<String?>(null)
+    val onchainReceiveAddress: StateFlow<String?> get() = _onchainReceiveAddress
+
+    private val _lastReceiveTxid = MutableStateFlow<String?>(null)
+    val lastReceiveTxid: StateFlow<String?> get() = _lastReceiveTxid
+
 
     private val _spendableOnchainSats = MutableStateFlow(0L)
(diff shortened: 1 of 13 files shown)


toneloc:
Not mergeable yet: the PR head doesn’t compile because `HomeScreen` passes `onClick` to `StatusCapsule`, but `StatusCapsule` doesn’t expose that parameter.

Even after that fix, the explorer-link change needs another pass because the channel-close branch now points at `lastRxTxid`, which can be an unrelated deposit txid; keep the new resolver for deposits only and port the proper close-txid approach separately.

The tooltip/spacing polish is worth keeping, but the resolver lifecycle should be reworked so it resolves while the pending-deposit card is active and clears stale txids on confirmation.

DevAnuragT:
> Not mergeable yet: the PR head doesn’t compile because `HomeScreen` passes `onClick` to `StatusCapsule`, but `StatusCapsule` doesn’t expose that parameter.
> 
> Even after that fix, the explorer-link change needs another pass because the channel-close branch now points at `lastRxTxid`, which can be an unrelated deposit txid; keep the new resolver for deposits only and port the proper close-txid approach separately.
> 
> The tooltip/spacing polish is worth keeping, but the resolver lifecycle should be reworked so it resolves while the pending-deposit card is active and clears stale txids on confirmation.

Pushed the fixes. The StatusCapsule compilation error is fixed, and the explorer link is now hidden for channel closes (we can port the proper close-txid approach separately). Also fixed a small UI race condition where the deposit link would briefly flash on startup after a channel close. Should be good to merge now.

toneloc:
Looks mostly good, but I’d hold merge on one issue: `startPendingDepositPolling()` now waits for `OnchainTxidResolver.resolve()` before entering the 10s `refreshBalances()` loop. Since resolver backoff can take several minutes or fail on explorer issues, confirmed onchain funds can stay stale for too long.

I’d run txid resolution separately/non-blocking, or interleave it with the existing balance polling. Minor: wrap OkHttp `execute()` responses in `.use {}` to avoid leaking response bodies.

What do you think?

DevAnuragT:
> Looks mostly good, but I’d hold merge on one issue: `startPendingDepositPolling()` now waits for `OnchainTxidResolver.resolve()` before entering the 10s `refreshBalances()` loop. Since resolver backoff can take several minutes or fail on explorer issues, confirmed onchain funds can stay stale for too long.
> 
> I’d run txid resolution separately/non-blocking, or interleave it with the existing balance polling. Minor: wrap OkHttp `execute()` responses in `.use {}` to avoid leaking response bodies.
> 
> What do you think?

yeah for the blocking issue: I moved the txid resolution into a separate background launch {} block so it doesn't block the 10-second balance polling loop. Also wrapped the OkHttp execute() call in a .use {} block to prevent leaks. I will push the fixes in sometime

DevAnuragT:
> Looks mostly good, but I’d hold merge on one issue: `startPendingDepositPolling()` now waits for `OnchainTxidResolver.resolve()` before entering the 10s `refreshBalances()` loop. Since resolver backoff can take several minutes or fail on explorer issues, confirmed onchain funds can stay stale for too long.
> 
> I’d run txid resolution separately/non-blocking, or interleave it with the existing balance polling. Minor: wrap OkHttp `execute()` responses in `.use {}` to avoid leaking response bodies.
> 
> What do you think?

I have addressed this, just getting explorer link error, tried 2,3 times, some or the other bug is coming, i have fixed them, till try the fixed one and push new commit

DevAnuragT:
@toneloc i have tested, the view explorer link is working fine now, also if u tap on total balance and change it to BTC format, then USD and btc balance below slider will change to BTC format, i think this way it's better

toneloc:
Thank you for this contrbution.

PR #183, opened 2026-07-15
refactor(ios): extract txid resolution and transaction link state

### LINKED ISSUE #182: iOS: Stale onchain transaction link persisting in the UI when generating a new receiving address
### Summary
When a user generates a new onchain receiving address (e.g. on the Fund Wallet or Receive screens), any previous `receiveTxid` stored in `AppState` is not cleared. This results in the UI incorrectly displaying a stale onchain transaction link (from a previous deposit) as if it corresponds to the newly generated address.

### Repro
1. Open the app and generate an onchain receive address.
2. Send an onchain payment to that address so a transaction link appears in the UI.
3. Once resolved or completed, generate a *new* onchain receive address.
4. Observe the UI: it still shows the transaction link from the *previous* deposit alongside the new address.

### Expected
Generating a new receiving address should immediately clear any pending/active receive transaction links in the UI, ensuring the user only sees a transaction link when an actual payment is detected on their *current* address.

alienx5499:
## Summary

Architectural refactor of the transaction link resolution and tracking system. Breaks down the bloated `AppState.swift` class to adhere strictly to the Single Responsibility Principle, separating UI state from background network polling and database operations.

---

## Problem

The original `AppState.swift` acted as a "God Object" for the iOS application. It managed the entire lifecycle of Esplora background resolvers, the `TxidLinkStore`, the SQLite database updates for payments, and exposed the UI state for transaction links directly. This led to:
- A massive, hard-to-read `AppState` class.
- Tangled dependencies between network polling, database writes, and UI state.
- Inconsistent and scattered replay logic.
- Leaking of internal storage implementation details to SwiftUI views.
- **Bug**: Stale onchain transaction links persisting in the UI when generating a new receiving address, causing confusion.

---

## Solution

1. **TransactionLinkService for UI State**: Extracted all transaction link UI state (such as `txidLinks` and `onchainReceiveAddress`) into a lightweight, `@Observable` service. It provides strict getters and setters to encapsulate the internal `TxidLinkStore` and has no external dependencies.
2. **TxidResolutionService for Network Polling**: Created a dedicated service to orchestrate the lifecycle of Esplora resolvers (`CloseTxidResolver`, `OnchainTxidResolver`) and their staggered launchers. It strictly manages polling and replay logic, and exposes closures (`didResolveTxid`) with a scoped `Event` enum.
3. **AppState as Composition Root**: Refactored `AppState` to coordinate these services. It configures the `TxidResolutionService`, intercepts resolution events, and wires the results into `DatabaseService` to record payments, while updating the `TransactionLinkService` UI state.
   - **Bug Fix (Stale UI Links)**: To prevent stale transaction links from persisting in the UI when a new deposit arrives, we now actively call `transactionLinkService.clearReceiveTxid()` inside `detectOnchainDeposit` the moment a balance increase is detected. This replaces the old, buggy approach of clearing the link whenever the user tapped "Receive", which was inadvertently wiping out the UI for actively confirming deposits.
4. **View Encapsulation**: Updated all SwiftUI views (`HomeView`, `ChannelSettingsView`, `FundWalletView`) to cleanly read from `appState.transactionLinkService` instead of piercing into internal `AppState` dictionaries.

---

## Screen Recording

https://github.com/user-attachments/assets/ac10d99e-c140-4236-96a8-d3ac533a1732

---

## Realted
- Closes #182


toneloc:
The refactor is mostly clean but there are a few blockers:

1. `resetInMemoryWalletState()` clears close/receive txids but does not clear `transactionLinkService.onchainReceiveAddress`, so a wallet reset can leave the old wallet’s receive address in memory and make later deposit detection resolve against the wrong address.

2. The PR description says `onchainReceiveAddress` has a `didSet` that clears the receive txid when a new address is generated, but `TransactionLinkService` currently has only a bare `var onchainReceiveAddress: String?`. Either add the `didSet` or fix the PR description.

3. `detectOnchainDeposit()` clears `lastReceiveTxid` before the `depositSats >= 1000` guard, so a tiny ignored balance fluctuation can erase a valid explorer link. Move the clear after the dust guard.

4. `TxidResolutionService.configureResolvers()` uses a one-time `precondition`; if startup configures resolvers and then node start fails, tapping “Try Again” calls `start()` again and can crash. Make configuration idempotent or clear the service on the startup failure path.

After those fixes, this looks mergeable.

alienx5499:
Thanks for the review, @toneloc! I’ve addressed all the points.

Regarding the `didSet` on onchainReceiveAddress, I had actually removed it earlier and updated the PR description, so you might have been looking at an older version. We kept it as a plain stored variable to avoid any observability side effects.

toneloc:
Thanks for these fixes.

PR #186, opened 2026-07-17
feat(ios): navigate to payment detail from home status bubble

### LINKED ISSUE #185: iOS: Home Payment Bubble Opens Payment Detail
### Summary

On the Home tab, the received-payment status bubble displays the latest received payment (amount, type, status) but is not interactive — tapping it does nothing. Users who receive a payment and want to see the full details (txid, confirmations, fee, counterparty) must manually navigate to the History tab and find the payment row themselves. This adds unnecessary friction to the most common post-payment action.

Additionally, the SQL query powering the bubble loads 10 rows and filters by `direction == "received"` in Swift memory — wasteful work when only one row is needed, and slower than a targeted index-friendly query.

### Why This Matters

- **Friction**: Receiving a payment is a key moment in the user journey. Tapping the bubble to see details is the expected iOS pattern (like tapping a notification banner). Forcing users to switch tabs and scroll breaks that flow.
- **Performance**: Loading 10 rows to find one is wasteful on every app foreground, especially as payment history grows.
- **Consistency**: The History tab already has payment rows that open `PaymentDetailView`. The Home bubble should behave the same way.

### Repro

1. Receive a payment (lightning, stability, splice, or onchain)
2. Open the app — Home tab shows a received-payment bubble with the amount
3. Tap the bubble — nothing happens
4. Manually switch to History → Payments and find the row to see details

### Expected Behavior

- Tapping the Home status bubble opens `PaymentDetailView` for the latest received payment
- The app switches to the History tab, Payments segment, with the detail sheet presented
- Dismissing the sheet leaves the user on the History tab — no forced return to Home
- Tapping a payment row in History also opens the same detail sheet
- The underlying SQL query fetches a single row instead of 10

alienx5499:
## Summary

Wires the Home status bubble to open the payment detail sheet, replaces the wasteful `getRecentPayments(limit: 10)` + in-memory filter with a focused `latestReceivedPayment()` SQL query, and fixes the tab/segment auto-switching behavior so users land on the correct segment when arriving from Home and stay there when dismissing the sheet.

---

## Problem

1. The Home tab's received-payment status bubble is non-interactive — tapping it does nothing.
2. `AppState.refreshLatestPaymentStatus` calls `getRecentPayments(limit: 10)` and filters by `direction == "received"` in Swift — wasteful work for a single-row result.
3. `DatabaseService` had no `latestReceivedPayment()` query, forcing callers to use the wider list-query.

## Solution

1. **New `latestReceivedPayment()` query** in `DatabaseService.swift` — single-row query with `WHERE direction = "received"` and the existing lightning-noise filter, ordered by `created_at DESC LIMIT 1`.
2. **`paymentRecord(from:)` helper** — extracted the 14-field `PaymentRecord` constructor from `getRecentPayments` into a private mapper so both queries share the same row-to-model code.
3. **`PaymentDetailCoordinator`** — pure navigation coordinator (`@Observable`) owning `selectedPayment`. Created in `MainTabView` and injected via `@Environment` into Home and History views. No DB logic.
4. **HomeView status bubble** — wrapped in a `Button` that calls `paymentCoordinator.open(payment)`.
5. **HistoryView payments row** — uses the same coordinator instead of a local `@State`, ensuring a single source of truth for the sheet.
6. **Tab & segment switching** — when `coordinator.selectedPayment` is set, the tab switches to History; the segment auto-switches based on `paymentType`. Dismissing the sheet leaves the user on History (no auto-return to Home).

---

## Screen Recording

https://github.com/user-attachments/assets/19706ae6-fc75-43c0-a330-7924efae107c

---

## Related
- Closes #185

### DIFF
--- a/ios/StableChannels/StableChannels.xcodeproj/project.pbxproj
+++ b/ios/StableChannels/StableChannels.xcodeproj/project.pbxproj
@@ -21,6 +21,7 @@
 		37DA6E9D4F5D4BF36DD86BC8 /* ReceiveView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 445B52C4DEC5CE8D3B0B3DA9 /* ReceiveView.swift */; };
 		3DBEBB9A611388A35C38F1D2 /* MainTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 623F41D841B312F0C9AE01E6 /* MainTabView.swift */; };
 		3E46529F1181F2F866A0A495 /* DatabaseService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3940DC5406ABC63AA090441A /* DatabaseService.swift */; };
+		491D2E883008F96B0009FADF /* PaymentDetailCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 491D2E873008F96B0009FADF /* PaymentDetailCoordinator.swift */; };
 		4923802D2FB51864007D9419 /* InvoiceScanView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4923802C2FB51864007D9419 /* InvoiceScanView.swift */; };
 		496A44432FCD4F75005AF12F /* QRInputToolbar.swift in Sources */ = {isa = PBXBuildFile; fileRef = 496A44422FCD4F75005AF12F /* QRInputToolbar.swift */; };
 		496A44442FCD4F75005AF12F /* QRCodeExtractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 496A44412FCD4F75005AF12F /* QRCodeExtractor.swift */; };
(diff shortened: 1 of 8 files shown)


toneloc:
Looks good overall, but I think this needs one fix before merge: `MainTabView` and `HistoryView` both present `PaymentDetailView` from the same `coordinator.selectedPayment` binding. When a payment is selected and the tab switches to History, both sheets can become active, causing duplicate/flaky presentation. I’d keep the sheet at `MainTabView` and remove the payment sheet from `HistoryView`; history rows can still call `paymentCoordinator.open(payment)`.

Please double-check this.

alienx5499:
You're right that presenting from both MainTabView and HistoryView would cause a race on the same coordinator binding. Fixed in 36c20d8: removed the sheet from HistoryView entirely. History rows now call paymentCoordinator.open(payment) the same way Home does, and MainTabView's .onChange handles the tab switch while presenting the sheet above the TabView hierarchy. Only one presenter, no duplicate sheet risk.

toneloc:

PR #190, opened 2026-07-20
feat(ios): event-driven on-chain confirmation tracking with local calculation & UI progress

### LINKED ISSUE #189: iOS: On-Chain Payment Confirmation Tracking & Event-Driven Engine
## Description

The iOS app previously treated on-chain transactions (`onchain`, `splice_in`, `splice_out`, `channel_close`) as instantly completed or left them stuck in static states without real-time tracking of blockchain confirmations. Users had no visual indication of transaction depth (e.g. `2/6 confirmed`), and continuous API polling created unnecessary network and database overhead.

This issue introduces a full-featured, event-driven confirmation tracking system on iOS:
1. An **event-driven Confirmation Polling Engine** triggered by blockchain tip changes (`BlockHeightService.onHeightUpdated`).
2. **Local confirmation calculation** using stored `tx_block_height` to eliminate redundant Esplora API network requests once a transaction is mined.
3. **Database write deduplication** to eliminate redundant SQLite `UPDATE` statements when confirmation count hasn't changed.
4. **Rich UI progress indicators** in `HistoryView` payment list rows (`2/6 confirmed`) and `PaymentDetailView` (`ProgressView` bar + green `Confirmed` badge).
5. **Clean database persistence** capping stored confirmations at `ConfirmationPolicy.requiredConfirmations` (6).

alienx5499:
## Summary

Implements event-driven on-chain payment confirmation tracking for iOS. Confirmation resolution is triggered directly by block height updates (`BlockHeightService.onHeightUpdated`), calculates confirmation progress locally once `tx_block_height` is mined to eliminate Esplora network hits, deduplicates SQLite writes, and surfaces real-time progress (`2/6 confirmed` labels and `ProgressView` bars) prominently to users.

---

## Problem

1. **No Confirmation Awareness**: On-chain, splice-in/out, and channel close transactions were treated as instantly final or static without reflecting live network progress.
2. **Network & Battery Overhead**: Naive polling hit external Esplora endpoints repeatedly even for transactions whose mining block height was already known.
3. **Database Churn**: SQLite writes occurred on every cycle even when the confirmation state had not changed.
4. **UI Gap**: Payment history rows showed only static `"Pending"` or `"Completed"` text without intermediate confirmation counts.

---

## Solution

1. **Event-Driven Architecture**: Wired `BlockHeightService.onHeightUpdated` to `ConfirmationPollingService.pollOnce()`, replacing uncoordinated polling loops with blockchain-driven updates.
2. **Local Calculation Engine**: In `ConfirmationService`, when `txBlockHeight` is already present, confirmation depth is computed locally (`currentHeight - txBlockHeight + 1`), reducing Esplora status calls to zero post-mining.
3. **DB Write Deduplication**: `ConfirmationPollingService` skips SQLite `UPDATE` execution when `progress.display == payment.confirmations`.
4. **Capped Persistence & Query Safety**: Stored `confirmations` in SQLite are capped at `ConfirmationPolicy.requiredConfirmations` (6) and SQL queries parameterize `requiredConfirmations` dynamically.
5. **Rich UI Progress**:
   - `PaymentRowView`: Displays `"2/6 confirmed"` in blue for 1–5 confs and `"Confirmed"` in green for 6+ confs.
   - `PaymentDetailView`: Displays interactive `ProgressView` bar for active confirmations positioned prominently in the main `Payment Details` section.
   - Preserved original iOS grouped inset card styling.
   - Event-driven reloads on block height updates (`.onChange(of: appState.blockHeightService.currentHeight)`).
---
## Architecture Diagram
image

---
## Screen Recording

https://github.com/user-attachments/assets/ca9eb288-c626-49df-887d-b0c96df2cf48

---

## Related

- Closes #189


alienx5499:
Updated the layout based on review feedback: moved the Confirmations progress row higher into the main `Payment Details` section (right under `Status`) for maximum visibility when opening the detail sheet!

image

toneloc:
Thanks for the update. The remaining blockers I’d want addressed before merge:

1. `/tx/{txid}/status` treats `confirmed: false` as a retry/miss, so a normal mempool tx can burn through backoff and stall the sequential confirmation poll for later payments.
2. `ConfirmationPollingService.onUpdate` is never wired, so History rows won’t live-refresh after DB confirmation updates. Detail refreshes on height changes, but can race before the poller writes.
3. `PaymentDetailView` still falls back to stored status at 0 confs, so a row can show `Completed` while the progress row says `0/6`.
4. Cached `tx_block_height` is never revalidated/updated, so shallow reorgs or re-mined txs can count toward 6 confs from stale height.
5. `ResilientEsploraClient.run()` now treats cancellation like exhaustion because `fetch()` returns `nil` for both, contradicting the documented silent-cancel contract.

alienx5499:
Hi @toneloc, pushed a commit addressing the points you raised. The unconfirmed tx parser no longer returns nil (which was triggering retries), so mempool transactions won't stall the polling queue. Wired up a confirmation update epoch so both HistoryView and PaymentDetailView reload after the DB write completes instead of racing on raw block height. Status labels at 0 confs now consistently show "0/6 confirmed" for on-chain types. Payments under 6 confs re-fetch from Esplora each block to catch any reorgs, and the cancel contract on ResilientEsploraClient is fixed.

toneloc:
This PR does not fully address reorgs yet. A block reorg is when Bitcoin temporarily replaces the latest block(s) with a different chain, so a transaction that was at height `100` can move to `101` or even become unconfirmed again.

The PR re-fetches the block height for payments under 6 confirmations, but the corrected height is not always saved because unchanged confirmation counts skip the DB write, and `COALESCE(tx_block_height, ?)` preserves the old height.

Fix by updating the DB when either the displayed confirmation count or `tx_block_height` changes. Since `paymentsNeedingConfirmation()` only selects rows below 6 confirmations, `updateConfirmations` can replace `COALESCE(tx_block_height, ?)` with `tx_block_height = ?`.

Hope this makes sense. Almost there.

toneloc:

PR #191, opened 2026-07-21
feat: Logs & Diagnostics UI for Android, iOS, and Desktop

DevAnuragT:
## Summary
This PR introduces a cross-platform Logs & Diagnostics feature for the Android, iOS, and Desktop apps, so users can export their logs (and we can use them to diagnose issues like unexplained force-closes) directly from the app.

## Details
- **Android:** Added `LogsView.kt` in the Settings page with Share/Download capabilities, a `DisclaimerBanner`, and a `FileProvider` (`res/xml/file_paths.xml`) to securely export ZIP files.
- **iOS:** Added a native `DiagnosticsSettingsView` into the `SettingsView` Support section, exporting the existing audit + LDK logs.
- **Desktop/Core:** `user.rs` generates the log ZIP archive natively in Rust (audit log + `ldk_node.log`).
- **Force-close diagnosability:** channel closes on all three platforms now record a structured `ClosureReason` (kind + peer message / processing error / payment hash, etc.) and the counterparty node id in the audit log, instead of an opaque debug string, so force-closes are easier to root-cause.
- **Android fix:** sharing logs launches the system share sheet, which used to background the app and trigger a full node stop/resync on return (visibly refreshing the UI). A transient flag now makes `MainActivity` skip that one background/foreground cycle when returning from the share sheet.
- Removed the unused `AppLogger`/`tracing` scaffolding added earlier in this branch, per review feedback, in favor of exporting the logs the app already writes.

## Testing
- Desktop: `cargo test` (includes new unit tests for the `ClosureReason` -> JSON mapping).
- Android: built and installed on a physical device (OnePlus, Android 15); verified log export and the share-sheet no-resync fix.
- iOS: built (signed) and run on simulator; verified log export.


ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+                    .fill(color)
+                    .frame(width: 32, height: 32)
+                Image(systemName: icon)
+                    .foregroundStyle(.white)
+                    .font(.system(size: 16, weight: .semibold))
+            }
+            Text(text)
+                .foregroundStyle(.primary)
+        }
+    }
```
alienx5499:
```suggestion

```

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+                    Button {} label: {
+                        rowLabel(icon: "square.and.arrow.up", color: .gray, text: "Share the logs")
+                    }.disabled(true)
+                    Button {} label: {
+                        rowLabel(icon: "arrow.down.doc", color: .gray, text: "Download logs")
+                    }.disabled(true)
+                }
+            } header: {
+                Text("Logs")
+            }
```
alienx5499, line 312:
```suggestion
            } footer: {
                Text("Save app logs to a file for debugging and support.")
            }
```

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+struct DiagnosticsSettingsView: View {
+    @State private var isExporting = false
+    @State private var logDocument: LogTextDocument? = nil
+
+    var body: some View {
+        List {
+            Section {
+                Text("Save app logs to a file for debugging and support.")
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
```
alienx5499:
```suggestion

```

alienx5499 (changes requested):
Hey @DevAnuragT, thanks for adding Logs & Diagnostics feature to the iOS side! 

I only reviewed the iOS UI part of the code here. There are a few small nitpicks regarding the layout and styling to make it match the rest of the app perfectly. I've left some inline code suggestions.

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+
+                let logUrls = AppLogger.shared.getExportLogs()
+                if !logUrls.isEmpty {
+                    ShareLink(items: logUrls) {
+                        rowLabel(
+                            icon: "square.and.arrow.up",
+                            color: .green,
+                            text: "Share the logs"
+                        )
```
alienx5499:
```suggestion
                        HStack {
                            Image(systemName: "square.and.arrow.up")
                                .foregroundStyle(.green)
                            Text("Share the logs")
                                .foregroundStyle(.primary)
                        }
```

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+                            }
+                        }
+                        logDocument = LogTextDocument(text: allLogs)
+                        isExporting = true
+                    } label: {
+                        rowLabel(
+                            icon: "arrow.down.doc",
+                            color: .green,
+                            text: "Download logs"
+                        )
```
alienx5499:
```suggestion
                        HStack {
                            Image(systemName: "arrow.down.doc")
                                .foregroundStyle(.green)
                            Text("Download logs")
                                .foregroundStyle(.primary)
                        }
```

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+                    }
+                } else {
+                    Button {} label: {
+                        rowLabel(icon: "square.and.arrow.up", color: .gray, text: "Share the logs")
```
alienx5499:
```suggestion
                        HStack {
                            Image(systemName: "square.and.arrow.up")
                            Text("Share the logs")
                        }
                        .foregroundStyle(.gray)
```

ios/StableChannels/StableChannels/Features/Settings/SettingsView.swift
```diff
@@ -239,3 +256,103 @@ struct SettingsView: View {
...
+                    }.disabled(true)
+                    Button {} label: {
+                        rowLabel(icon: "arrow.down.doc", color: .gray, text: "Download logs")
```
alienx5499:
```suggestion
                        HStack {
                            Image(systemName: "arrow.down.doc")
                            Text("Download logs")
                        }
                        .foregroundStyle(.gray)
```

alienx5499 (changes requested):

toneloc:
This PR is not mergeable yet because it breaks builds. `src/user.rs` still contains unresolved conflict content around `=======` / `>>>>>>>`, and `export_logs_to_zip()` appears to be left inside the test module instead of module scope, so the desktop app cannot compile. 

Android also references `@xml/file_paths` from `AndroidManifest.xml`, but the PR does not add `res/xml/file_paths.xml`, so the FileProvider resource is missing.

For this PR, can we remove the new `AppLogger`/tracing approach and keep the scope to exporting the logs the app already writes today, like `audit_log.txt` and the LDK node logs? The new logger is not wired through the real app paths yet, so we can add an app-wide logger later in a separate PR instead of shipping unused scaffolding here.

How does this sound?

DevAnuragT:
> This PR is not mergeable yet because it breaks builds. `src/user.rs` still contains unresolved conflict content around `=======` / `>>>>>>>`, and `export_logs_to_zip()` appears to be left inside the test module instead of module scope, so the desktop app cannot compile.
> 
> Android also references `@xml/file_paths` from `AndroidManifest.xml`, but the PR does not add `res/xml/file_paths.xml`, so the FileProvider resource is missing.
> 
> For this PR, can we remove the new `AppLogger`/tracing approach and keep the scope to exporting the logs the app already writes today, like `audit_log.txt` and the LDK node logs? The new logger is not wired through the real app paths yet, so we can add an app-wide logger later in a separate PR instead of shipping unused scaffolding here.
> 
> How does this sound?

Agreed on dropping the AppLogger/tracing approach — removed on all three platforms, and scoped to exporting the logs we already write audit_log.txt. The conflict markers and the export_logs_to_zip placement are fixed (desktop compiles), and I'll add the missing res/xml/file_paths.xml for the FileProvider.

Since the aim is diagnosing force-closes, I'd also like to tighten the existing CHANNEL_CLOSED audit event to record the full ClosureReason peer_msg instead of the current toString/Debug form — matching what the LSP already does.

DevAnuragT:
@alienx5499 I have updated files, u may check them now

toneloc:
Looking better, thanks.

Please asses these two issues for accuracy:

1. [Download writes directly to public storage](https://github.com/toneloc/stable-channels/blob/c00d817db1225359c90ae82abc0e8d3568fb2d8a/android/app/src/main/java/com/stablechannels/app/services/LogExporter.kt#L74-L92). This can fail under Android scoped storage. Please use `MediaStore.Downloads` or the system file picker.

2. [Sharing suppresses the entire background lifecycle](https://github.com/toneloc/stable-channels/blob/c00d817db1225359c90ae82abc0e8d3568fb2d8a/android/app/src/main/java/com/stablechannels/app/MainActivity.kt#L63-L84). If the user continues into another app, the node remains active and the eventual foreground sync is skipped. Please remove this suppression or scope it safely to the chooser.

DevAnuragT:
> Looking better, thanks.
> 
> Please asses these two issues for accuracy:
> 
> 1. [Download writes directly to public storage](https://github.com/toneloc/stable-channels/blob/c00d817db1225359c90ae82abc0e8d3568fb2d8a/android/app/src/main/java/com/stablechannels/app/services/LogExporter.kt#L74-L92). This can fail under Android scoped storage. Please use `MediaStore.Downloads` or the system file picker.
> 2. [Sharing suppresses the entire background lifecycle](https://github.com/toneloc/stable-channels/blob/c00d817db1225359c90ae82abc0e8d3568fb2d8a/android/app/src/main/java/com/stablechannels/app/MainActivity.kt#L63-L84). If the user continues into another app, the node remains active and the eventual foreground sync is skipped. Please remove this suppression or scope it safely to the chooser.

implemented

toneloc:

PR #201, opened 2026-07-28
android: custom LSP configuration

DevAnuragT:
## Summary

Adds a "Custom LSP Configuration" feature to the Android app so it's no longer hardcoded to `stablechannels.com`. Users can point the app at their own LSPS2-capable LSP, with validation, a safety gate against switching while channels are open, and automatic rollback if a new LSP config fails to start **or connect**.

## Changes

- **`LspPreferencesManager.kt`** (new): SharedPreferences-backed manager storing an optional custom LSP pubkey/address, falling back to `Constants.DEFAULT_LSP_PUBKEY`/`DEFAULT_LSP_ADDRESS` when unset. Includes format validation (`isValidPubkey`, `isValidAddress`) and `saveCustomLsp`/`resetToDefault`.
- **`NodeService.kt`, `StabilityProcessingService.kt`, `AppState.kt`**: all hardcoded `Constants.DEFAULT_LSP_*` reads replaced with `LspPreferencesManager` lookups, so both the foreground node build and the background stability service pick up the active (possibly custom) LSP.
- **`StableChannel.kt`**: added `defaultWithLsp(context)` factory so runtime defaults resolve the LSP pubkey dynamically; the static `DEFAULT` is kept for compile-time/test use.
- **`AppState.switchLsp()` / `resetLspToDefault()`**: validate + persist, then perform an in-process `NodeService.stop()`/`start()` restart (gated on `channels.isEmpty()`), reusing the already-hardened stop/start + `LdkNodeOwner` lifecycle rather than a full app-process restart.
- **`LspSettingsView.kt`** (new) + `SettingsHub.kt`/`SettingsNavigation.kt`: new "LSP" settings screen — shows active address/node ID, disables editing with a warning while channels are open, "Switch LSP" dialog with validation, and "Reset".

## Node restart approach

LDK bakes the LSP pubkey/address into `Config`, `AnchorChannelsConfig`, and the LSPS2 liquidity source at node-build time, so a plain reconnect isn't enough — the node has to be rebuilt. This is only ever attempted with **no open channels**, which makes an in-process `NodeService.stop()` + `start()` the safest option: it reuses the existing, already-tested stop/start lifecycle (including `LdkNodeOwner` release/reacquire) without tearing down the Activity, ViewModel, or other background jobs the way a full app-process restart via `Intent` would.

## Safety hardening

- **Auto-rollback on bad config**: LDK validates the pubkey is a real secp256k1 point at build time (format-only regex isn't sufficient — roughly half of arbitrary hex strings that pass the 66-char/`02`/`03` check are *not* valid curve points). If the node fails to restart with a newly-saved LSP, `switchLsp()` automatically reverts to the previously-working config and restarts again, so a bad save never leaves the node stopped.
- **Rollback on connect failure**: `NodeService.start()` propagates a failed `connect()` for a custom LSP so an unreachable/mismatched host triggers the rollback path instead of being saved as "successful" (mirrors the merged iOS `NodeService.swift`).
- **Background-service race guard**: `performLspNodeRestart()` calls `waitForBackgroundService()` before touching `LdkNodeOwner`, and uses `cancelAndJoin()` (not `cancel()`) for the periodic stability/heartbeat jobs so they're guaranteed stopped before the node is torn down and rebuilt.
- **Authoritative counterparty**: `sc.counterparty` is refreshed from the LSP pref on switch and derived from the live channel's `counterpartyNodeId` on every balance refresh, so trades/keysends always target the node the channel is actually with (the `channels` table doesn't persist the counterparty).

## Review feedback addressed

- **Remove "Default" wording** → badge shows "Custom" only; button is now "Reset". (`fa54d7e`)
- **#1 stale in-memory counterparty after switch** → set from the LSP pref on switch when no channel exists, and derive the authoritative value from the live channel's `counterpartyNodeId` on every balance refresh (matches the iOS PR #207 "restore counterparty from matching channel" fix). (`fa54d7e`, `a8094be`)
- **#2 swallowed connect failures** → `NodeService.start()` rethrows a failed `connect()` when a custom LSP is set, so an unreachable/refused LSP triggers rollback instead of saving a broken config as successful. (`fa54d7e`)
- **#3 reset lacked rollback** → `resetLspToDefault()` now captures the prior custom config and restores + restarts on failure. (`fa54d7e`)
- Dialog copy aligned with the wording approved on iOS. (`a8094be`)

## Updates

- **Follow-up review comments** (`a61fef6`): the channels-active gate now requires the node to be running and cross-checks both the live channel list and persisted DB state (new `DatabaseService.hasAnyChannel()`) before allowing a switch — a stopped/mid-restart node previously reported an empty channel list and could slip through. Connect-failure strictness is now scoped to only the LSP-switch restart path (`NodeService.start(strictLspConnect = true)`); ordinary cold starts / foreground resumes log-and-continue on a connect failure instead of treating it as fatal, so a transient LSP outage doesn't block normal startup.
- **Found during on-device retesting** (`e6c0ce2`): if the rollback's own restart also failed — reproduced this with a `FeerateEstimationUpdateFailed` during a slow-network window — the node was left stopped with no automatic recovery until a manual relaunch. `switchLsp()`/`resetLspToDefault()` now fall back to the app's existing `scheduleNodeStartRetry()` mechanism in that case instead of leaving the node down indefinitely.

## Testing

Compiled and unit-tested with JDK 17 (`:app:compileDebugKotlin`, `:app:testDebugUnitTest`), then verified live on a physical device:

| | |
|---|---|
| New "LSP" row in Settings hub | Settings hub |
| LSP settings screen (default state) | LSP settings default |
| Switch LSP dialog validation | Validation error |
| Auto-rollback after a bad LSP save | Auto rollback |
| Reset restores the default LSP | Reset |

On-device verification of the review fixes:
- **Unreachable custom LSP → rollback**: a valid-curve pubkey with an unreachable `203.0.113.9:9735` produced `LSP connect failed: Network connection closed` → `Failed to restart node after LSP switch — rolling back` → node recovered on default, UI showed *"Invalid LSP — reverted to previous settings."*
- **Successful switch** to a reachable LSP shows the "Custom" badge and re-enables **Reset**; **Reset** restores default and restarts cleanly (node re-registers with the LSP).
- **Channels-active gating** verified with a real open channel: the config is hidden and the warning *"Your LSP cannot be changed while channels are active…"* is shown.
- **Hardened gate + retry fallback**: re-verified the channels-active gate re-enables/disables correctly across opening and closing a real channel; reproduced a double-failure (bad pubkey → rollback's own `Node.start()` failing on a transient network blip) twice, confirmed the node was left stopped before the retry-fallback fix, and confirmed after the fix the app still recovers to a running, correctly-reverted default LSP (via the existing `scheduleNodeStartRetry()` path).

### Not yet covered
- A genuinely working *second* LSPS2 LSP end-to-end (would require switching the hardcoded `Network.BITCOIN` to signet/mutinynet to test against `server/stable-channels-lsp` — out of scope here).
- The counterparty-derive *mismatch-correction* branch can't be forced through the UI because the channels-active gate keeps the LSP config and the open channel's counterparty in sync during normal use; it's a defensive measure covered by code review and verified to run correctly with a live channel.


toneloc:
Thanks for this.

Let's please remove the word "Default" - we can keep the "Reset" button. 

Additonally, please check the comments below for accuracy:

1. [Switching only restarts the node (AppState.kt L566–584)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/AppState.kt#L566-L584); it never updates the in-memory channel counterparty. [Trades continue using that stale pubkey (TradeService.kt L18–65)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/services/TradeService.kt#L18-L65), so the first channel after switching can still communicate with the previous LSP.

2. [Connection failures are swallowed (NodeService.kt L143–148)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/services/NodeService.kt#L143-L148). An unreachable or mismatched LSP is saved as successful, so rollback never runs. Verify peer/LSPS2 readiness with a timeout.

3. [Reset to Default lacks rollback (AppState.kt L541–560)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/AppState.kt#L541-L560). If restart fails, the previous working custom configuration has already been removed.

DevAnuragT:
> Thanks for this.
> 
> Let's please remove the word "Default" - we can keep the "Reset" button.
> 
> Additonally, please check the comments below for accuracy:
> 
> 1. [Switching only restarts the node (AppState.kt L566–584)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/AppState.kt#L566-L584); it never updates the in-memory channel counterparty. [Trades continue using that stale pubkey (TradeService.kt L18–65)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/services/TradeService.kt#L18-L65), so the first channel after switching can still communicate with the previous LSP.
> 2. [Connection failures are swallowed (NodeService.kt L143–148)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/services/NodeService.kt#L143-L148). An unreachable or mismatched LSP is saved as successful, so rollback never runs. Verify peer/LSPS2 readiness with a timeout.
> 3. [Reset to Default lacks rollback (AppState.kt L541–560)](https://github.com/toneloc/stable-channels/blob/d863160eafaa011dd00d02783f727593df69f95b/android/app/src/main/java/com/stablechannels/app/AppState.kt#L541-L560). If restart fails, the previous working custom configuration has already been removed.

All three addressed + "Default" wording removed. #1: counterparty now synced from the LSP pref on switch and derived from the live channel's `counterpartyNodeId` (matches the iOS #207 fix). #2: `NodeService.start()` rethrows a failed `connect()` for custom LSPs so unreachable hosts roll back instead of saving as successful. #3: `resetLspToDefault()` now captures and restores the prior config on failure. Verified on-device (incl. unreachable-LSP rollback); PR description updated.

toneloc:
Thanks; please check these further comments for accuracy.

1. **LSP connect failure bricks every node start for custom-LSP users** — `NodeService.start()` now rethrows a failed `connect()` whenever a custom LSP is configured; the outer catch stops the node and the app lands in `Phase.ERROR`, not watching the chain. Transient LSP unreachability at cold start or foreground resume disables the app. Scope the strictness to the LSP-switch restart path only (e.g. a `strictLspConnect` flag set by `performLspNodeRestart()`) and let ordinary starts log-and-continue.
https://github.com/toneloc/stable-channels/blob/a8094be52a23c89567d957e5fcd327d3b751594a/android/app/src/main/java/com/stablechannels/app/services/NodeService.kt#L143-L153

2. **No-open-channels gate is bypassable when the node isn't running** — `refreshChannels()` returns `node?.listChannels() ?: emptyList()`, so the gate sees an empty list whenever the node is stopped or mid-restart; a user with a live channel on disk can then switch LSPs, breaking the counterparty relationship the channel depends on. Require `nodeService.isRunning` before the gate and consult persisted channel state as a second source of truth.
https://github.com/toneloc/stable-channels/blob/a8094be52a23c89567d957e5fcd327d3b751594a/android/app/src/main/java/com/stablechannels/app/AppState.kt#L509-L513

DevAnuragT:
> Thanks; please check these further comments for accuracy.
> 
> 1. **LSP connect failure bricks every node start for custom-LSP users** — `NodeService.start()` now rethrows a failed `connect()` whenever a custom LSP is configured; the outer catch stops the node and the app lands in `Phase.ERROR`, not watching the chain. Transient LSP unreachability at cold start or foreground resume disables the app. Scope the strictness to the LSP-switch restart path only (e.g. a `strictLspConnect` flag set by `performLspNodeRestart()`) and let ordinary starts log-and-continue.
>    https://github.com/toneloc/stable-channels/blob/a8094be52a23c89567d957e5fcd327d3b751594a/android/app/src/main/java/com/stablechannels/app/services/NodeService.kt#L143-L153
> 2. **No-open-channels gate is bypassable when the node isn't running** — `refreshChannels()` returns `node?.listChannels() ?: emptyList()`, so the gate sees an empty list whenever the node is stopped or mid-restart; a user with a live channel on disk can then switch LSPs, breaking the counterparty relationship the channel depends on. Require `nodeService.isRunning` before the gate and consult persisted channel state as a second source of truth.
>    https://github.com/toneloc/stable-channels/blob/a8094be52a23c89567d957e5fcd327d3b751594a/android/app/src/main/java/com/stablechannels/app/AppState.kt#L509-L513

Implemented

toneloc:
Looks good now. Thanks for this contribution.

PR #203, opened 2026-07-28
feat(iOS): SPV Header Chain

### LINKED ISSUE #202: iOS: Lightweight SPV Header Chain & Reorg Detection Engine
### Summary

Introduces a local SPV (Simplified Payment Verification) header chain service to ingest Bitcoin block headers, track chain tip linkage via `previousblockhash`, detect chain reorganizations (reorgs), and automatically roll back orphaned headers and confirmed payments in SQLite.

### Why This Matters

- **Reorg Safety**: On-chain payment confirmations can be invalidated if Bitcoin experiences a blockchain reorganization. Without tracking header linkages, orphaned payments remain marked as confirmed.
- **Header Chain Integrity**: Validating `previousblockhash` linkage ensures the local wallet follows the canonical chain.
- **Bounded Persistence**: Storing full headers indefinitely bloats mobile database storage. A rolling 2016-block window (1 difficulty epoch, ~2 weeks) bounds storage to ~160 KB.
- **Event-Driven Execution**: Interlocking WebSocket block announcements with SPV validation guarantees instant reorg detection and payment re-verification.

### Expected Behavior

- Mempool WebSocket block header stream feeds into `SPVHeaderChainService`.
- Sequential headers append to local `block_headers` SQLite table.
- Reconnect duplicates are dropped idempotently via `headerExists`.
- Offline gaps (app backgrounded, height jump > 1) update tip without false reorg rollbacks.
- Chain reorgs (prevHash mismatch at same/lower height) locate common ancestor, purge orphaned headers, and reset affected payments to `pending` status in an atomic SQLite transaction.
- Headers older than 2016 blocks are automatically pruned.

alienx5499:
## Summary

Introduces `SPVHeaderChainService` to ingest Bitcoin block headers streamed from Mempool WebSocket, persist them in a local SQLite rolling window, detect chain reorganizations (reorgs), and execute atomic database rollbacks of orphaned block headers and confirmed payments.

---

## Problem

1. **Reorg Vulnerability**: Previous confirmation tracking assumed linear, monotonically increasing block heights. A Bitcoin chain reorganization could orphan mined blocks, leaving payments in SQLite marked as `completed` with stale `tx_block_height`s.
2. **Lack of Header Linkage**: Without storing block headers and verifying `previousblockhash` linkages, the wallet could not cryptographically identify fork points or common ancestors during chain splits.
3. **Unbounded Storage**: Indefinitely accumulating block headers in mobile SQLite causes storage bloat over time.
4. **Redundant Polling & Double-Execution**: Chain updates during reorg resolution could trigger duplicate polling loops and un-coalesced database writes.

---

## Solution

1. **`SPVHeaderChainService.swift`** — Core state machine for SPV header management:
   - Ingests headers containing `height`, `id` (hash), `previousblockhash`, and `timestamp`.
   - Sequential extension: Appends header to SQLite when `prevHash == tip.hash`.
   - Idempotency guard: Skips processing via `headerExists(height:hash:)` when exact header is already stored (WS reconnection).
   - Offline gap handling: Distinguishes backgrounding gap jumps (`incomingHeight > tip.height + 1`) from reorg forks, advancing tip without false rollbacks.
   - Reorg resolution: When `prevHash` mismatches tip, searches backward for the common ancestor, rolls back orphaned headers and payments above common ancestor, and plants the new canonical tip.
2. **Rolling Window Header Pruning (`DatabaseService.swift`)** — Automatically prunes headers older than 2016 blocks (1 difficulty epoch, ~2 weeks). Keeps table size bounded to ~160 KB on disk.
3. **Atomic SQLite Transactions & Correct Rollback Target**:
   - `rollbackPaymentsConfirmedAfter(height:)`: Targets `tx_block_height` with `IS NOT NULL` guard (fixing timestamp mismatch), resetting status to `pending`, `confirmations = 0`, and `tx_block_height = NULL`.
   - Wrapped header deletion, payment status reversion, and new tip insertion inside `DatabaseService.inTransaction` to prevent partial state inconsistencies.
4. **Optimized Height Updates (`BlockHeightService.swift`)**:
   - Added `setHeightSilently(_:)` to update height counter without firing `onHeightUpdated`, eliminating redundant duplicate `pollOnce()` cycles during reorg processing.
5. **Confirmation Subdirectory Reorganization**:
   - Consolidated confirmation and height tracking components into `Services/Confirmation/` directory.

---

## Related
- Closes #202


toneloc:
Thanks for this contribution.

Please check the below for accuracy. 

The main case to fix before merge is an offline reorg. A height gap is currently treated as harmless, while fully confirmed payments are never rechecked. That can leave a reversed payment marked completed permanently.

For an early-stage solution, don’t build a full header-walking implementation. When a gap or conflicting header appears:

1. Refresh the authoritative tip through Esplora.
2. Revalidate recently confirmed onchain payments—including completed payments from the last ~12 blocks.
3. Downgrade any transaction Esplora reports as unconfirmed.
4. Add tests for an ordinary offline gap and a fully confirmed payment orphaned during that gap.

The WebSocket should trigger revalidation, not directly decide which payments to roll back. The deeper SPV validation, stale-block guard, TLS pinning, and uncommon lower-height edge cases can be follow-ups.

toneloc:
Thanks for this. Please review this for accuracy:

1. **Deferred `BEGIN` in the shared app/NSE database — use `BEGIN IMMEDIATE`** — the new `inTransaction` opens with plain `BEGIN;` and `handleReorg` reads before writing: the classic deferred read→write lock upgrade that can fail with SQLITE_BUSY against the NSE and abort the reorg rollback. Every other transaction in this codebase uses `BEGIN IMMEDIATE` for exactly this reason.
https://github.com/toneloc/stable-channels/blob/bcf41b362ae9fc9b4e56090b7cd42f58d37fe0ef/ios/StableChannels/StableChannels/Services/DatabaseService.swift#L1414-L1425

2. **WS-supplied height now drives confirmation counting directly** — previously the WS block event only triggered an Esplora HTTP refresh; now `advanceHeight(to: block.height)` pushes the raw mempool.space value into `BlockHeightService`, so an inflated feed value lets the cached-height fast path compute fake 6+ confs and flip payments to `completed` with no HTTP check — and the later poll won't un-complete them. Clamp WS heights against the HTTP-verified tip, or keep `refresh()` as the height authority.
https://github.com/toneloc/stable-channels/blob/bcf41b362ae9fc9b4e56090b7cd42f58d37fe0ef/ios/StableChannels/StableChannels/Services/Confirmation/SPVHeaderChainService.swift#L183-L190

3. **Corrupt header store crashes on every incoming block** — `fetchLatestHeader` uses trapping `UInt32(height)`/`UInt32(timestamp)` conversions on `Int64` DB values; one out-of-range row means a crash loop with no recovery. Use `UInt32(exactly:)` and treat a bad row as an empty store — this is disposable cache data.
https://github.com/toneloc/stable-channels/blob/bcf41b362ae9fc9b4e56090b7cd42f58d37fe0ef/ios/StableChannels/StableChannels/Services/DatabaseService.swift#L1547-L1563

alienx5499:
Thanks @toneloc! Updated as requested and rebased this PR directly on top of database PR #209 (`refactor/ios-database-repositories`):

1. **`BEGIN IMMEDIATE` Transactions**: Updated `inTransaction` in `DatabaseService.swift` to use `mode: "IMMEDIATE"` (`BEGIN IMMEDIATE;`), preventing read-to-write lock upgrade failures and `SQLITE_BUSY` errors against the NSE.
2. **Esplora Height Authority**: Updated `advanceHeight` in `SPVHeaderChainService.swift` to trigger an authoritative Esplora HTTP `refresh()`, ensuring height updates and confirmation counts are always driven by HTTP-verified tips rather than raw WebSocket feed values.
3. **Safe Header Decoding**: Replaced trapping `UInt32(...)` conversions in `fetchLatestHeader` with `UInt32(exactly:)` to safely handle corrupt/out-of-range values without crashing.

Please merge PR #209 first, and then review this PR. Ready for review!

toneloc:
Thanks for this . The last commit (1e01cac) introduced two regressions that block merge, plus one revalidation gap:

1. **App DB filename changed, splitting the app from the NSE and from existing installs** — `DatabaseService` now opens `stable_channels.sqlite`, while main, the README, and `NotificationService` still use `stablechannels.db`. Existing users would boot into a fresh empty DB, and the NSE would keep using the old file, so shared rows like `pending_stability_send` stop being shared. Restore the original filename — this looks like a fragment of the pre-trim #209 that shouldn't have come back.
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/StableChannels/Services/DatabaseService.swift#L9-L10
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/NotificationService/NotificationService.swift#L330-L339

2. **The BEGIN IMMEDIATE fix got regressed** — 76546d2 fixed this, but the new `DatabaseService.inTransaction(mode: String = "DEFERRED")` wrapper defaults to deferred and `handleReorg` calls it without a mode, so the reorg rollback is back to `BEGIN DEFERRED` — the same shared app/NSE read→write lock-upgrade hazard as before. Make this path `BEGIN IMMEDIATE` (and consider defaulting the wrapper to IMMEDIATE so this can't regress silently again).
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/StableChannels/Services/DatabaseService.swift#L225-L229
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/StableChannels/Services/Confirmation/SPVHeaderChainService.swift#L154-L187

3. **Completed payments can stay completed after a reorg drops them below 6 confs** — revalidation only downgrades when Esplora reports the tx unconfirmed; if it re-confirmed in a newer block with fewer confirmations, `updateConfirmations` lowers the count but its `CASE WHEN confs >= required THEN 'completed' ELSE status END` never demotes status. Make status follow confirmations during revalidation (completed → pending when confs < required).
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/StableChannels/Services/Confirmation/ConfirmationPollingService.swift#L101-L107
https://github.com/toneloc/stable-channels/blob/1e01cac849410946c8fab19a619bf87a14b842d7/ios/StableChannels/StableChannels/Services/Repositories/PaymentRepository.swift#L167-L178

toneloc:
Thanks for these fixes.

PR #192, opened 2026-07-22
Umbrel Stable Channels LSP app-store package 

toneloc:
Self-contained Umbrel Community App Store package for the Stable Channels LSP:
- umbrel/ — app manifest, app_proxy compose, hooks/pre-start config rendering, three production Dockerfiles (ldk-server / sc-lsp / lsp-gui), the LSPS2 restart self-healing guard, and the tier-1 local sim + test store server.
- .github/workflows/umbrel-images.yml — multi-arch GHCR image publish.
- server/ — the shared LSP GUI + REST-client changes the package builds from: zero-input auto-connect via /setup/key.txt with exponential-backoff retry (lsp-server-gui), origin-relative REST (sc-rest-client).

Branched from main so it can be reviewed/submitted independently of the e2e harness and desktop work on autotest.

Claude-Session: https://claude.ai/code/session_017bVpHkcmJUnWegH8vAHZ2e

PR #210, opened 2026-07-31
Refine Umbrel packaging

Billa05:
Builds on #192 and completes the Umbrel integration by:
- Fixing WASM connectivity through Umbrel’s reverse proxy and TLS setup.
- Preserving operator-edited configuration across restarts, including optional mobile push settings.
- Fixing LDK/audit log handling and clipboard support over local HTTP.
- Adding safer container shutdown, port handling, and a real community-store testing workflow.

Tested locally with Umbrel dev and Bitcoin Core regtest.


toneloc:
Thanks for this. Made one small change. 

As mentioned: current image tags are placeholders for the pre-release flow; after merging, will rebuild the Umbrel images from the merge commit and update docker-compose.yml to pin the published image@sha256:... digests.

PR #211, opened 2026-08-01
fix(android): reliable photo-picker QR scanning + background-stop ANR fix

DevAnuragT:
## Summary

Makes the Send screen's **"import QR from photo library"** feature actually work reliably. While testing the existing photo-picker path on a real device (Oppo/ColorOS), it failed in three distinct ways — this PR fixes all of them.

### 1. Every picked image failed with "Could not read image"
The image was decoded via `InputImage.fromFilePath(context, uri)`, which opens the `content://` stream twice (bounds/EXIF, then pixels) and throws `FileNotFoundException` on some OEM providers for photo-picker URIs. The replacement decode path also had a bug where it used the return value of `BitmapFactory.decodeStream(..., inJustDecodeBounds = true)` to detect success — but that call **always returns `null` by design**, so it rejected every image.

**Fix:** a single-stream-open `decodeSampledBitmap()` that detects success from the stream opening + decoded `outWidth`/`outHeight`, downsamples oversized photos (`inSampleSize`) to avoid `OutOfMemoryError`, and hands a `Bitmap` to `InputImage.fromBitmap(...)`.

### 2. App got ANR-killed / dropped to "Syncing wallet" when opening the picker
Opening the system photo picker briefly backgrounds the app, firing `MainActivity.onPause()` → `AppState.stopNodeForBackground()`, which called the **blocking native LDK `node.stop()` on the main thread**. That blocked the UI thread for >5s, so Android raised an ANR (`Waited 5001ms for FocusEvent`) and killed the process; on relaunch the node had to fully resync.

**Fix:**
- The immediate node-stop now runs on `Dispatchers.IO` (tracked via `backgroundStopJob`, consistent with the existing 60s grace path), so `onPause()` returns instantly and no longer ANRs.
- Added an `isPickingMedia` guard so the transient `onPause` caused by the in-app picker skips the node teardown entirely — no more "Syncing wallet" screen on return.

### 3. State leaks & undifferentiated errors in the scan callback
`isExtractingQR` was reset in three separate callbacks (risking a stuck spinner), failures were swallowed silently (`catch (_: Exception)`), and every failure mode showed the same "No Lightning invoice or Bitcoin address QR code was found" toast.

**Fix:** the scan is now a single coroutine with one `finally { isExtractingQR = false }`, failures are logged via `android.util.Log`, and the user sees a differentiated message:
- "Could not read the selected image" — image unreadable
- "No QR code was found in that image" — ML Kit ran but found nothing
- "QR code found, but it's not a Lightning invoice or Bitcoin address" — QR decoded but not a payment string

## Note on scope
The `AppState` changes (async node-stop + `isPickingMedia`) are slightly broader than the picker itself, but they're the reason the picker was unusable (ANR + forced resync), so they're bundled here.

## Testing
Verified end-to-end on a physical Oppo `CPH2531` (Android 15 / ColorOS):
- Picking a valid Lightning-invoice QR extracts the invoice into the field.
- No ANR / crash / "Syncing wallet" when opening and returning from the picker.
- Unreadable images, QR-less images, and non-payment QR codes each show the correct message.

## Files
- `android/app/src/main/java/com/stablechannels/app/ui/transfer/SendScreen.kt`
- `android/app/src/main/java/com/stablechannels/app/AppState.kt`

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -394,10 +394,7 @@ class AppState(private val context: Context) : ViewModel() {
             Log.d("AppState", "Stopping node immediately (no active payment request)")
             // node.stop() is a blocking native call; run it off the main thread so onPause()
             // returns immediately and Android doesn't ANR-kill us on the focus-change timeout.
-            backgroundStopJob?.cancel()
-            backgroundStopJob = viewModelScope.launch(Dispatchers.IO) {
-                performBackgroundStop()
-            }
+            launchBackgroundStop()
             return
         }
 
@@ -415,10 +412,24 @@ class AppState(private val context: Context) : ViewModel() {
             Log.e("AppState", "Failed to start LdkBackgroundService", e)
         }
 
-        backgroundStopJob = viewModelScope.launch(Dispatchers.IO) {
-            delay(60000L) // 60 seconds delay
-            performBackgroundStop()
+        launchBackgroundStop(delayMs = 60000L)
+    }
+
+    private fun launchBackgroundStop(delayMs: Long = 0L) {
+        backgroundStopJob?.cancel()
+        val job = viewModelScope.launch(Dispatchers.IO) {
+            try {
+                if (delayMs > 0L) {
+                    delay(delayMs)
+                }
+                performBackgroundStop()
+            } finally {
+                if (backgroundStopJob === coroutineContext[Job]) {
+                    backgroundStopJob = null
+                }
+            }
         }
+        backgroundStopJob = job
     }
 
     fun cancelBackgroundStop() {
@@ -454,7 +465,6 @@ class AppState(private val context: Context) : ViewModel() {
     }
 
     private fun performBackgroundStop() {
-        backgroundStopJob = null
         try {
             LdkBackgroundService.stop(context)
         } catch (e: Exception) {
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -66,6 +66,11 @@ class AppState(private val context: Context) : ViewModel() {
     @Volatile
     var isWaitingForPayment = false
 
+    // Set while an in-app system picker (e.g. photo picker) is open, so the transient onPause
+    // it triggers doesn't tear down and resync the LDK node.
+    @Volatile
+    var isPickingMedia = false
+
     private val _errorMessage = MutableStateFlow("")
     val errorMessage: StateFlow<String> = _errorMessage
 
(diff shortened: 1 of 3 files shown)


toneloc:
Thanks for this.

Please check these below comments for accuracy:

1. **Async immediate stop races `restartNodeFromForeground()`; can leave WALLET phase with a dead node** — the stop now runs on `Dispatchers.IO`, and `cancelBackgroundStop()` cannot interrupt it once the blocking, non-suspending `nodeService.stop()` has begun. A quick pause→resume can observe `isRunning == true`, take the "still running, reconnecting" path, then have the in-flight stop kill the node afterward — UI stays in `Phase.WALLET` with payments failing. Serialize with `cancelAndJoin()` before the `isRunning` check, or a shared `Mutex` around stop and restart.
https://github.com/toneloc/stable-channels/blob/69d14dd97ff4c508ab1149fdf6d56fc9c947b895/android/app/src/main/java/com/stablechannels/app/AppState.kt#L382-L390

2. **`isPickingMedia` is unbounded suppression with a stuck-true failure mode** — the flag is set before `photoPickerLauncher.launch()` and cleared only in the launcher callback; if `launch()` throws or the composition is disposed (unregistering the launcher), it stays true for the process lifetime and every later background transition skips the node stop entirely. Route picker-open pauses into the existing 60 s grace path instead of skipping the stop, set the flag only after `launch()` returns, and clear it defensively from `onResume`/`DisposableEffect`.
https://github.com/toneloc/stable-channels/blob/69d14dd97ff4c508ab1149fdf6d56fc9c947b895/android/app/src/main/java/com/stablechannels/app/ui/transfer/SendScreen.kt#L356-L361

DevAnuragT:
> Thanks for this.
> 
> Please check these below comments for accuracy:
> 
> 1. **Async immediate stop races `restartNodeFromForeground()`; can leave WALLET phase with a dead node** — the stop now runs on `Dispatchers.IO`, and `cancelBackgroundStop()` cannot interrupt it once the blocking, non-suspending `nodeService.stop()` has begun. A quick pause→resume can observe `isRunning == true`, take the "still running, reconnecting" path, then have the in-flight stop kill the node afterward — UI stays in `Phase.WALLET` with payments failing. Serialize with `cancelAndJoin()` before the `isRunning` check, or a shared `Mutex` around stop and restart.
>    https://github.com/toneloc/stable-channels/blob/69d14dd97ff4c508ab1149fdf6d56fc9c947b895/android/app/src/main/java/com/stablechannels/app/AppState.kt#L382-L390
> 2. **`isPickingMedia` is unbounded suppression with a stuck-true failure mode** — the flag is set before `photoPickerLauncher.launch()` and cleared only in the launcher callback; if `launch()` throws or the composition is disposed (unregistering the launcher), it stays true for the process lifetime and every later background transition skips the node stop entirely. Route picker-open pauses into the existing 60 s grace path instead of skipping the stop, set the flag only after `launch()` returns, and clear it defensively from `onResume`/`DisposableEffect`.
>    https://github.com/toneloc/stable-channels/blob/69d14dd97ff4c508ab1149fdf6d56fc9c947b895/android/app/src/main/java/com/stablechannels/app/ui/transfer/SendScreen.kt#L356-L361

Both fixed in latest commit: 1. restartNodeFromForeground now cancelAndJoin()s the pending stop before checking isRunning, 2. isPickingMedia routes through the existing 60s grace path instead of skipping the stop, set after launch() returns, cleared defensively in onResume/onDispose. Tested on-device (fast pick: no stop/no sync; slow pick >60s: grace timer fires and reconnects normally).

toneloc:
Thanks for this contribution.

PR #217, opened 2026-08-04
add signed trade rejections, correlation, and expiry for #205

### LINKED ISSUE #205: Add signed trade rejection and pending-trade expiry
Scope:
- LSP sends a signed NACK when TRADE_V1 validation fails.
- Include a reliable trade/payment correlation identifier and rejection reason.
- Wallets mark the corresponding trade failed instead of leaving it pending.
- Expire trades that never receive an accepted SYNC_V1.
- Persist expiry/rejection across restarts.
- Implement consistently on desktop, Android, and iOS.

Billa05:
closes #205. Implements signed trade outcomes, durable correlation, and desktop expiry. Automatic fee refunds remain intentionally deferred.

## What changed

  - Added optional signed trade_id and payment correlation fields.
  - Added signed trade rejections with stable reason codes.
  - Desktop now waits for authoritative LSP acceptance and displays sending, awaiting, completed, rejected, failed, and expired states.
  - Added 15-minute desktop expiry with authoritative late-outcome handling.
  - Persisted decisions atomically, deduplicated replays, and added durable response retries with dead-lettering after 168 attempts.
  - Preserved legacy desktop and mobile compatibility.

  ## Rejection conditions

  - `invalid_amount`:  invalid expected_usd or malformed trade_id.
  - `stale_request`: timestamp is more than 15 minutes old or ahead.
  - `invalid_fee`: fee inputs are invalid or the received fee does not match the signed trade.
  - `invalid_quote`: missing/invalid timestamp, quote, or LSP price.
  - `quote_out_of_range`: wallet quote differs from the LSP price by more than 1%.
  - `invalid_allocation`: missing quote/backing pair, backing exceeds available sats, zero allocation is inconsistent, or backing differs from the USD target by over $0.25.
  - `insufficient_capacity`:  requested USD allocation exceeds channel capacity.
  - `duplicate_trade`: another payment already used the same trade_id.
  - `internal_error`: required state, signing, database, or persistence operation failed.

  Invalid signatures, incorrect channel binding, and unattributable payments are dropped without an authenticated rejection.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ bitcoin = "0.32"
 ureq = { version = "2.10.1", features = ["json"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+rand = "0.9"
 retry = "1.3"
 futures = "0.3"
 async-trait = "0.1"
@@ -67,7 +68,6 @@ long_description = "A self-custodial bitcoin wallet with USD stability."
 tempfile = "3"
 electrsd = { version = "0.36.1", features = ["legacy", "esplora_a33e97e1", "corepc-node_27_2"] }
 electrum-client = "0.24.0"
-rand = "0.9"
 
 [workspace]
 members = [
@@ -82,5 +82,4 @@ members = [
 # macOS / Linux builds.
 [target.'cfg(target_os = "windows")'.build-dependencies]
 winres = "0.1"
-
 
--- a/server/stable-channels-lsp/src/event_loop.rs
+++ b/server/stable-channels-lsp/src/event_loop.rs
@@ -258,11 +258,28 @@ async fn dispatch(
             let direction = e.payment.as_ref().map(|p| if p.direction == 1 { "outbound" } else { "inbound" });
             let mut settlement_handled = false;
             if let Some(payment_id) = payment_id.as_deref() {
-                let known_settlement = state
-                    .db
-                    .settlement_exists(payment_id)
-                    .ok()
-                    .unwrap_or(false);
+                match state.db.mark_trade_response_succeeded(payment_id) {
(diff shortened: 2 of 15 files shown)


toneloc:
Thanks for this. Please check these two merge-blocking issues for accuracy:

### Ordinary `SYNC_V1` messages fail the new channel-ID validation

`src/user.rs` now requires every `SYNC_V1` message’s `user_channel_id` to equal the wallet’s local LDK ID. However, `send_sync_message` in `stable_manager.rs` populates ordinary syncs with the LSP’s local ID, and the two IDs normally differ. Only correlated trade responses echo the wallet’s ID.

Consequently, ordinary syncs—including stability backstops, spend deductions, startup flushes, and splice reconciliation—are dropped as `SYNC_V1_CHANNEL_MISMATCH`, causing wallet and LSP accounting to diverge.

Please either validate ordinary syncs using the shared `channel_id` and reserve the wallet-local `user_channel_id` check for correlated trade responses, or maintain an explicit remote-ID mapping. An end-to-end test passing a normal `build_sync_payload` message through the wallet validation would cover this regression.

### A persistence failure after keysend causes unbounded duplicate sends

`send_pending_trade_response` performs the real keysend before calling `mark_trade_response_in_flight`. If that database update fails, the obligation remains pending and immediately due, while `response_attempts` and `next_response_attempt_at` remain unchanged.

The one-second retry worker will therefore send another real keysend every tick indefinitely. Because the attempt counter never advances, the obligation can never reach the dead-letter threshold.

Please durably reserve/increment the attempt and apply backoff before sending, and do not send if that transition fails. Also add a failure-injection test where `spontaneous_send` succeeds but the subsequent persistence step fails, asserting that no immediate duplicate send occurs.

PR #221, opened 2026-08-08
derive stable-channel trade backing locally on both sides

Billa05:
- Each peer derives backing using its own BTC price and channel balances.
- The wallet quote is used only for fee/slippage validation.
- expected_usd remains the shared agreed value.
- Peer-supplied backing fields are ignored.
- TRADE_V1 payments are explicitly displayed as trades, distinct from SYNC_V1.

### DIFF
--- a/server/lsp-server-gui/src/state.rs
+++ b/server/lsp-server-gui/src/state.rs
@@ -62,6 +62,7 @@ pub enum DisplayUnit {
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub enum SettlementKind {
 	Stability,
+	Trade,
 	Sync,
 }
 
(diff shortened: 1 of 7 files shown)


toneloc:
Thanks for this contribution.

Merging this now. The trust-model change is correct: peer-supplied backing is no longer used for accounting. The boundary fixes in `74c12d6` and `456f595` address the review regressions.

Suggested follow-ups, in priority order:

### 1. Delta-based trade accounting

Trades still reprice the entire position, which can erase unsettled stability drift. This behavior already exists on main.

Fix both wallet and LSP together:

- Sign the USD delta, old target, and `trade_id`.
- Each peer converts the delta to sats using its own price.
- Apply the delta to current backing instead of storing an absolute allocation.
- Keep re-anchoring when a real stability payment settles.

This may cause a stability payment after a trade when unsettled drift remains, which is the correct behavior.

### 2. Explicit trade correlation

Trade acknowledgments currently match only on `expected_usd`, so duplicate targets can complete the wrong trade.

Add a client-generated `trade_id` to `TRADE_V1` and echo it in the acknowledging `SYNC_V1`. Define rollout behavior for older peers that do not echo the ID.

PR #224, opened 2026-08-09
fix(ios): remove the NSE wait flag and bound chain resolution

toneloc:
waitForNSE polled an `nse_processing` UserDefaults flag that the NSE sets but cannot reliably clear: the extension runs a full LDK node inside a ~24MB / ~30s budget, and when iOS jetsams it, cleanup() never runs and the flag stays set permanently. Every subsequent launch then burned the full 30s ceiling waiting on a dead process. Nine of nine NSE_WAIT events across three weeks of logs were pinned at exactly 30s with no partial waits — the app was never waiting for a live extension, only for a corpse.

NodeDirLock (flock on ldk-node.lock) already provides the same mutual exclusion and self-heals, because the kernel releases the lock when the holding process dies. Measured dir_lock_ms is 0-1ms and NODE_LOCK_TIMEOUT has never fired. The flag is removed from the NSE too, where it is now a dead write.

Also on the critical path, all of which the launch screen was waiting on:

- resolveChainURL used URLSession.shared, whose 60s default request timeout could hold the launch screen for a minute on a slow primary. It now has a 2s budget of its own.
- setLogPath ran inside initializeDatabaseServices, i.e. after the prologue it was meant to measure, so every startup audit event was silently dropped. Moved to the top of start().
- The prologue is now timed end to end (STARTUP_PROLOGUE), since chain resolution and the dir lock previously logged only on failure and were otherwise invisible.

Measured cold start: ~31s -> under 1s. Prologue 297-479ms, of which chain resolution is 283-468ms; LDK itself starts in 0.23s.

### DIFF
--- a/.github/workflows/swift-format.yml
+++ b/.github/workflows/swift-format.yml
@@ -26,7 +26,13 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install SwiftFormat
-        run: brew install swiftformat
+        # Pinned: the runner image ships a pre-installed older brew version, so
+        # `brew install` is a no-op and the effective linter version drifts with
+        # the image. The config uses 0.62-only rule names.
+        run: |
+          curl -sL -o /tmp/swiftformat.zip https://github.com/nicklockwood/SwiftFormat/releases/download/0.62.1/swiftformat.zip
+          unzip -o /tmp/swiftformat.zip -d /tmp/swiftformat-bin
+          echo "/tmp/swiftformat-bin" >> "$GITHUB_PATH"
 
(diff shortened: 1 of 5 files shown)


alienx5499 (approved):
LGTM

Verified the performance metrics side by side on simulators:
* **Startup time** is down from **~31 seconds** to **0.42 seconds** (resolving the thread block in `waitForNSE`).
* **Active runtime stats** (idle CPU, RAM, and thread stability) remain virtually identical and stable after launch, with no background resource regression.

Great optimization. Good to go!

PR #225, opened 2026-08-09
Preserve stability drift across trades

Billa05:
## Problem

  Trades repriced the entire stable position at the latest price, so even tiny or no-op trades could erase existing stability drift.

  Independent wallet/LSP prices could also disagree near channel capacity.

  ## Fix

  - Apply only the `expected_usd` delta at each peer’s local price.
  - Preserve existing drift and cumulative rounding.
  - Reject underflow and allocations above the live balance.
  - Validate locally before sending the fee.
  - Keep LSP capacity strict; `$0.25` remains only a settlement threshold.
  - Reserve the 0.5% quote-deviation band in the wallet.
  - Treat sub-cent targets as zero.
  - Require settlement before unsafe reductions or full exits.

  Both peers derive backing independently. No wire-format or schema changes.

  ## Trade-off

  Near-capacity trades may leave a small native BTC remainder. Exact `peg_all` negotiation is deferred.

  ## Follow-ups

  1. **Signed stability payments:** bind the channel, exact amount, direction, settlement ID, and expiry. This replaces the unsafe one-byte marker and prevents replay or incorrect settlement accounting.

  2. **Signed trade results:** add authorization, correlation, expiry, deduplication, retries, and `peg_all` negotiation. This prevents fees from leaving wallets stuck waiting for an uncorrelated sync.

### DIFF
--- a/server/stable-channels-lsp/src/stable_manager.rs
+++ b/server/stable-channels-lsp/src/stable_manager.rs
@@ -15,11 +15,10 @@ use ldk_server_client::ldk_server_grpc::api::{
 };
 use ldk_server_client::ldk_server_grpc::events::ChannelStateChangeReason;
 use ldk_server_client::ldk_server_grpc::types::{Channel, CustomTlvRecord};
+use stable_channels::constants::MAX_TRADE_QUOTE_DEVIATION_PERCENT;
 use stable_channels::db::Database;
 use stable_channels::types::{Bitcoin, StableChannel, USD};
 use tracing::{error, info};
-
-const MAX_TRADE_QUOTE_DEVIATION_PERCENT: f64 = 0.5;
 
 /// Return each peer's own spendable-plus-reserve balance from the fields LDK exposes for that
 /// peer. `channel_value - local_balance` is not the remote balance: on outbound channels it also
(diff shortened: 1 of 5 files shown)


toneloc:
Thanks for this. Merging but with these two comments.

**Reject zero backing for a live target.** `trade_backing_after_delta` can return zero backing for a nonzero target. `check_stability` then treats it as unset and rebuilds the backing, erasing drift. Reject this state or represent initialized zero separately. [Trade](https://github.com/toneloc/stable-channels/blob/f6c8627a47c217d3c87d42c776b5d9f146230019/src/stable.rs#L377-L435) · [Repair](https://github.com/toneloc/stable-channels/blob/f6c8627a47c217d3c87d42c776b5d9f146230019/src/stable.rs#L674-L682)

**Validate reductions across the accepted quote range.** The wallet applies the 0.5% safety band only to increases, so a reduction can pass locally but be rejected by the LSP after the fee is paid. Check reductions and exits across the full accepted price interval. [Wallet](https://github.com/toneloc/stable-channels/blob/f6c8627a47c217d3c87d42c776b5d9f146230019/src/user.rs#L10673-L10730) · [LSP](https://github.com/toneloc/stable-channels/blob/f6c8627a47c217d3c87d42c776b5d9f146230019/server/stable-channels-lsp/src/stable_manager.rs#L1880-L1950)

PR #222, opened 2026-08-08
feat(android): add mempool websocket tx tracking and receive flow stabilization

DevAnuragT:
## Summary
- add Android mempool websocket implementation for real-time onchain transaction detection
- add reconnect, dedup, and transaction matching support services for websocket event handling
- wire websocket receive and outspend events into app state and channel-close/onchain flows
- add websocket unit tests for service, matcher, reconnect, and dedup components
- include onchain confirmation-flow hardening and home/history/details UX consistency updates validated during websocket integration

## Main Changes
- new websocket services under android app services websocket package
- app state integration for receive, removed, and tracked-outspend handling
- confirmation polling and receive txid ownership/address-safety logic
- history and payment detail confirmation/explorer-link improvements
- home onchain section pending-receive visibility improvements

## Validation
- manual on-device receive and confirmation-state checks performed
- manual channel close tested
- manual splice in tested

PR #228, opened 2026-08-10
fix(iOS): align websocket receive tracking and reconciliation with android improvements

alienx5499:
## Summary

Ports the Mempool WebSocket tracking and receive flow reconciliation improvements from Android PR #222 to iOS, resolving phantom receive row writes on channel closures and ensuring address reuse safety during placeholder cleanup.

---

## Problem

1. **Phantom Receive Row Generation**: Spends of tracked transaction IDs (such as channel closures) triggered address-match events in the WebSocket receive path, resulting in false `.receive` broadcasts and empty phantom database entries.
2. **Resolution Cleanup Conflicts on Reused Addresses**: The fallback HTTP receive resolution path deleted pending placeholder rows based purely on the existence of a corresponding transaction ID (`txid`), without matching the payment amounts. When a user received multiple concurrent deposits to the same reused address, resolving one deposit would wipe out the placeholder row of another deposit, causing history loss.

---

## Solution

1. **Strict Receive Loop Filtering**:
   - Filtered out `match.isTxid` events in `MempoolWebSocketService.swift`'s address transaction matching loop. Tracked transaction events (such as channel close spends) are now processed solely by the outspends tracked transactions pipeline.
2. **Transaction ID Retrieval**:
   - Added `payment(txid:)` lookup capability to `PaymentRepository.swift` to retrieve recorded payment details by transaction hash.
3. **Amount-Gated Reconciliation**:
   - Updated `handleOnchainReceiveResolved(resolutionId:txid:)` in `AppState.swift` to retrieve the WebSocket payment and check `wsPayment.amountMsat == UInt64(row.amountMsat)`.
   - The pending placeholder is only deleted if the amounts match exactly. Otherwise, the placeholder is preserved as a separate deposit.

---

## Key Changes

1. **`MempoolWebSocketService.swift`**:
   - Excluded `isTxid` matches from triggering incoming receive events.
2. **`PaymentRepository.swift`**:
   - Added `payment(txid: String) -> PaymentRecord?` query.
3. **`AppState.swift`**:
   - Guarded duplicate placeholder deletions in `handleOnchainReceiveResolved` on exact `amountMsat` matching.
4. **`DatabaseServiceTests.swift`**:
   - Added `testPaymentByTxid()` to verify database retrieval.

### DIFF
--- a/ios/StableChannels/StableChannels/App/AppState.swift
+++ b/ios/StableChannels/StableChannels/App/AppState.swift
@@ -1831,10 +1831,22 @@ class AppState {
     private func handleOnchainReceiveResolved(resolutionId: Int64, txid: String) {
         if let db = databaseService,
            let row = db.onchainRepo.fetchPendingOnchainReceiveRow(resolutionId: resolutionId) {
-            if db.paymentRepo.paymentExists(txid: txid, excludePaymentId: row.paymentId) {
-                // WebSocket beat us to it, this fallback placeholder is a duplicate.
-                db.paymentRepo.deletePayment(paymentId: row.paymentId)
+            if let wsPayment = db.paymentRepo.payment(txid: txid) {
+                if wsPayment.amountMsat == UInt64(row.amountMsat) {
+                    // WebSocket beat us to it, this fallback placeholder is a duplicate.
+                    db.paymentRepo.deletePayment(paymentId: row.paymentId)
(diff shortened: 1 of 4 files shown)


toneloc:
Thanks for this contribution. Appreciate you learning from the Android improvements.

PR #233, opened 2026-08-11
authenticate and amount-bind stability settlements

Billa05:
## Problem

  Lightning authenticates the payment transport between nodes, but it does not tell our application that a custom TLV is authorized to change a specific stable channel’s accounting.

  The existing one-byte stability marker did not bind the payment’s sender, channel, amount, direction, or settlement identity. Trade and sync payloads are signed for the same reason: their application-level meaning must be authenticated separately from the Lightning payment carrying them.

  ## Solution

  This PR adds a signed stability-settlement envelope that binds:

  - Stable channel and settlement identity
  - Exact payment amount and direction
  - Expiry and replay protection

  The receiver verifies the stable counterparty’s node signature, matches the signed amount against the actual payment, and applies the settlement atomically using its own local price. Durable deduplication and retries make processing safe across crashes and transient failures. The LSP remains compatible with older mobile clients.

  This does not replace Lightning’s signatures. Lightning secures payment delivery; this signature authorizes the Stable Channels accounting instruction carried inside it.

  The PR also fixes a channel-ID representation bug that could leave valid settlements pending indefinitely, plus a trade edge case where zero backing for a non-zero target could later erase preserved stability drift.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ bitcoin = "0.32"
 ureq = { version = "2.10.1", features = ["json"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+rand = "0.9"
 retry = "1.3"
 futures = "0.3"
 async-trait = "0.1"
@@ -67,7 +68,6 @@ long_description = "A self-custodial bitcoin wallet with USD stability."
 tempfile = "3"
 electrsd = { version = "0.36.1", features = ["legacy", "esplora_a33e97e1", "corepc-node_27_2"] }
 electrum-client = "0.24.0"
-rand = "0.9"
 
 [workspace]
 members = [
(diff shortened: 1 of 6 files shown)


toneloc:
This looks okay, but I am worried that we are making changes that impact Android and iOS behavior as well so the server deployment gets more complicated.

Are the same changes for iOS and Android planned?

 Have other recent server updates impacted iOS and Android behavior, and/or is the logic of user.rs included in iOS and Android?

Billa05:
Every protocol change we are making is backward-compatible, so deploying the latest LSP will not interrupt existing mobile clients. I opened #235 to track removing these compatibility paths once Android and iOS are updated.

toneloc:
> Every protocol change we are making is backward-compatible, so deploying the latest LSP will not interrupt existing mobile clients. I opened #235 to track removing these compatibility paths once Android and iOS are updated.

Great, thanks for the clarification. LGTM.

PR #238, opened 2026-08-13
Add durable, signed LSP trade rejections(TRADE_REJECTED_V1)

Billa05:
## Problem

  When the LSP rejected a trade, the wallet received no result and could remain pending indefinitely.

  ## Solution

  - Return signed acceptance or rejection results.
  - Persist decisions and retry lost responses.
  - Safely deduplicate repeated trade requests.
  - Keep rejected trades from changing channel state.
  - Also removed the +-0.5% check on the wallet side, and added a warning for full peg trade that trade fees is non-refundable

### DIFF
--- a/server/stable-channels-lsp/src/event_loop.rs
+++ b/server/stable-channels-lsp/src/event_loop.rs
@@ -258,6 +258,24 @@ async fn dispatch(
             let direction = e.payment.as_ref().map(|p| if p.direction == 1 { "outbound" } else { "inbound" });
             let mut settlement_handled = false;
             if let Some(payment_id) = payment_id.as_deref() {
+                match state.db.mark_trade_response_delivered(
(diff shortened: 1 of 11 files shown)


toneloc:
Thanks I don't really see any blockers except for over-detailed stuff. 

How has this been tested? Also, iOS and Android will be ok, right?

Billa05:
I tested it on my regtest LSP deployed on AWS. I deliberately sent stale timestamps and mismatched prices and confirmed the LSP returned signed rejections, also tested the full-peg rejection flow.

Android and iOS remain compatible: they omit the optional trade_id, so the LSP keeps them on the existing TRADE_V1/SYNC_V1 flow.

toneloc:
Ok, thank you. LGTM.

PR #239, opened 2026-08-14
Expand price feeds across platforms

alienx5499:
## Summary

Expands the iOS Bitcoin price feed oracle from 5 feeds to 17 diverse global exchange and aggregator endpoints. Enhances `PriceService.swift` with generic JSON traversal capable of parsing top-level JSON arrays and nested array/dictionary structures with key-based filtering.

Replaces delayed merchant settlement rates with Bitcoin-native infrastructure (`Mempool.space`), achieving high fault tolerance against single-exchange outages while maintaining tight price convergence (0.064% delta, standard deviation of ±$5.16).

---

## Problem & Background

1. **Vulnerability to Outages & DNS Failures**: The previous 5-feed implementation was fragile. If 2 feeds experienced transient network timeouts or Cloudflare blocks, the median fell back to only 3 feeds, leaving the wallet vulnerable to price skew.
2. **Top-Level JSON Array Parsing Limitation**: Public APIs from several major exchanges (Bitfinex, Coinlore, Gate.io, and CoinDCX) return root-level JSON arrays (`[ ... ]`) rather than dictionaries (`{ ... }`). `PriceService.swift` previously cast raw responses strictly to `[String: Any]`, causing top-level array feeds to silently fail.
3. **Outlier Feeds & Delayed Merchant Rates**: Polling merchant conversion endpoints like BitPay introduced significant latency and built-in spreads ($600-$700 lower than live spot order books), skewing the median.

---

1. **Generic JSON Extraction & Root Array Support (`PriceService.swift`)**:
   - `extractPrice(from:path:)` accepts `Any` root objects and dynamically resolves numeric string keys as integer array indices.
   - `filterJSON(_:filterKey:filterValue:)` recursively searches both dictionary trees and arrays of objects to match target trading pairs (e.g. `market: "BTCUSDT"`).
   - Gracefully handles trailing value arrays (such as Kraken's `["<price>", "<vol>"]`).

2. **17-Feed Global Price Oracle (`Constants.swift`)**:
   - **Tier 1 Global Spot & USDT Order Books**: Binance, Kraken, Coinbase, Bitfinex, Bybit, KuCoin, Gate.io, MEXC, Huobi.
   - **Aggregators & Bitcoin Infrastructure**: CoinGecko, Blockchain.com, Coinlore, Yadio, Mempool.space.
   - **Regional Liquidity Providers**: Luno, CoinDCX, BTCTurk.

3. **High-Reliability Bitcoin-Native Price Source**:
   - Integrated `Mempool.space` (`/api/v1/prices`) as a pure USD fiat spot oracle, replacing delayed merchant settlement feeds.

4. **Concurrent Wall-Time Architecture**:
   - All 17 feeds execute in parallel using Swift concurrency (`withTaskGroup`). Total wall time is bounded by the slowest single feed (~800ms - 1.2s), incurring zero UX latency penalty compared to 5 feeds.

---

## Architecture Diagram

image

---

## Verification & Testing

   - Executed a 5-run benchmark with 15-second intervals comparing the 5-feed baseline against the 17-feed oracle across live market feeds.
   - **Average Delta**: +$40.01 (+0.064%) with standard deviation of ±$5.16.
   - **Fault Tolerance**: Maintained uninterrupted median consensus with 15+ feeds when individual feeds experienced network timeouts.

---

### DIFF
--- a/.github/workflows/android-tests.yml
+++ b/.github/workflows/android-tests.yml
@@ -1,0 +1,53 @@
+name: Mobile Tests – Android
+
+on:
+  pull_request:
+    paths:
+      - 'android/**'
+      - '.github/workflows/android-tests.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'android/**'
+      - '.github/workflows/android-tests.yml'
+
+concurrency:
+  group: test-android-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-android:
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: android
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+
+      - name: Set up Android SDK
+        uses: android-actions/setup-android@v3
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      # JVM unit tests (includes PriceOracleTest.kt — the cross-platform oracle parity suite).
+      - name: Run unit tests
+        run: |
+          chmod +x gradlew
+          ./gradlew --no-daemon testDebugUnitTest
+
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: android-unit-test-report
+          path: android/app/build/reports/tests/testDebugUnitTest
+          if-no-files-found: ignore
--- a/.github/workflows/ios-tests.yml
+++ b/.github/workflows/ios-tests.yml
@@ -18,8 +18,15 @@ concurrency:
 jobs:
   test:
     runs-on: macos-15
+    timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+
+      - name: Select Xcode 16.4
+        run: |
+          sudo xcode-select -s /Applications/Xcode_16.4.app/Contents/Developer
+          xcodebuild -version
+          xcrun simctl list devices available
 
       - name: Run iOS unit tests
         run: |
@@ -27,5 +34,6 @@ jobs:
           xcodebuild test \
             -project ios/StableChannels/StableChannels.xcodeproj \
             -scheme StableChannels \
-            -destination 'platform=iOS Simulator,OS=latest,name=iPhone 16' \
+            -destination 'platform=iOS Simulator,OS=18.5,name=iPhone 16' \
+            -parallel-testing-enabled NO \
             CODE_SIGNING_ALLOWED=NO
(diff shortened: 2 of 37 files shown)


toneloc:
Thanks working on this right now. It's a great idea and I think we should expand this PR to have Android and Desktop changes as well. 

One thing that I changed was the reliance on USDT (Tether). I would rather just use USD pairs because Tether can depeg. But, let's have it so we use USDT as a backup if several of the 6 USD pairs are not available.

I also ran some tests via a VPN of different countries. I actually could not test India!

## Geo-availability evidence for the 17-feed change (30 countries via Mullvad)

Tested every feed endpoint from 30 countries (each exit verified via `am.i.mullvad.net`), to check the geo-blocking / regional-outage concern behind this PR.

**Result: 23 of 30 countries returned all feeds.** Only two are genuinely reduced, both reproducible:

- **US (15/17):** Binance `451` + Bybit `403` — both geo-block US IPs (Binance.US is separate).
- **Turkey (12/17):** national filtering null-routes the 5 smaller non-CDN hosts (Yadio, Mempool, Luno, CoinDCX, BTCTurk); every big CDN-fronted exchange survived.

Everything else — all of Europe, LatAm, Asia-Pacific, Africa — came back clean. Scattered single-country `403`s on CoinGecko/CoinDCX/Mempool were **rate-limiting from hammering ~30 exits in 35 min**, not geo-blocks (each of those feeds worked from ~28 of the other 29 exits).

### On the resilience guards (all confirmed present)

The oracle now handles feed loss correctly, and the added tests pin it:
- **quorum of 3** — a single surviving feed can never set the price;
- **3s per-feed timeout** — a hung feed can't stall the whole update (matters most for high-latency regions);
- **10% median-move sanity band + $1000 floor**; USD-primary → USDT-fallback tiering.

### Caveats
- **India, Pakistan, Vietnam, Russia, Ethiopia, Bangladesh, South Korea, Yemen, Venezuela have no Mullvad exit** and could not be tested. India especially warrants a separate check (regulatory geo-blocks). Turkey (12/17) is the closest proxy for a heavy-filtering state.
- This expands only the **wallet's** feeds; the **LSP still runs its own USD feeds**, so the ±0.5% trade-band alignment between the two is unchanged by this PR.

alienx5499:
Tested all feed endpoints in India:

- **16 of 18 feeds returned 200 OK** (only Binance.US and Coinbase geo-restricted/errored as expected).
- **Direct USD Tier 1**: Bitstamp (~670ms), Kraken (~410ms), Blockchain.com (~430ms), Bitfinex (~1400ms), and Mempool.space (~1330ms) all resolve cleanly.
- **Aggregators & Regional**: CoinDCX resolves in ~87ms locally, plus CoinGecko, Binance, Bybit, KuCoin, Gate.io, MEXC, Huobi, Coinlore, Yadio, and BTCTurk are all healthy.
- Full iOS (207 tests) and Rust workspace tests are passing.

toneloc:
Looking good now. Testing this on-device now.

PR #240, opened 2026-08-17
add durable two-phase trade confirmation

Billa05:
This PR prevents users from paying a trade fee before knowing whether the LSP will accept the trade.

  Desktop trades now use a two-step flow. The wallet first sends a 1-msat proposal, and the LSP validates the quote, channel capacity and post-fee allocation before reserving the trade for 60 seconds. The wallet enables confirmation only after verifying the LSP’s signed response. Once confirmed, it sends the exact fee and the LSP atomically applies the reserved allocation.

  The reservation locks the approved price, fee and allocation, so later market movement cannot invalidate it. Refreshing creates a new quote at the current price, while cancelling releases the reservation. Expired confirmations cannot be executed, but payments that settled before expiry remain valid even if processed later.

  The flow is persisted across restarts, prevents duplicate execution, and keeps the final signed `SYNC_V1` authoritative. Full-peg trades use post-fee channel capacity without exceeding the available sats.

  Android, iOS and older wallets remain unchanged: trades without a `phase` continue through the existing fee-first flow.

### DIFF
--- a/server/lsp-server-gui/src/ui/payments.rs
+++ b/server/lsp-server-gui/src/ui/payments.rs
@@ -561,7 +561,12 @@ fn render_payment_details_dialog(ctx: &Context, app: &mut LspServerApp) {
 									.as_ref()
 									.map(|k| format_payment_kind(k))
 									.unwrap_or_else(|| "Unknown".to_string());
-								ui.label(payment_type);
+								let settlement_kind = app
+									.state
+									.settlement_kinds
+									.as_ref()
+									.and_then(|kinds| kinds.get(&payment.id).copied());
+								ui.label(payment_type_label_str(&payment_type, settlement_kind));
 								ui.end_row();
 
 								// Amount (unit-aware, pre-formatted above)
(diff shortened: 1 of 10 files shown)


toneloc:
I know we discussed this, but now that I look at the implementation, I’m not sure this is worth the complexity. 

Trades basically work today.

What does this give us beyond protecting users from the relatively rare case where the LSP rejects a trade after the fee has already settled? Does that benefit justify the two additional payments, reservation state, expiry handling, database migrations, and new failure modes? I am doubtful. What do you think @Billa05?

Billa05:
Yeah, I had the same thought, this is probably over-engineered. Its main value is ensuring users never lose a non-refundable trade fee when the LSP rejects a trade, which is most likely near a full peg. 

If we introduce refunds, that would probably have same implementation and recovery complexity as this anyway. If we’re comfortable leaving the case of an unrefunded fee for now, we can skip this.

toneloc:
Cool. Let's keep this on file for when / whether there are failed trades and the need to automate trade fee refunds.

PR #242, opened 2026-08-18
feat(iOS): implement automatic failover to secondary Esplora provider for node startup

### LINKED ISSUE #241: iOS: LDK Node Startup Failover on Feerate Estimation and Esplora Outage
### Summary

Implements automatic failover during LDK Node startup (`node.start()`) to gracefully recover from Esplora rate-limits, network timeouts, and `FeerateEstimationUpdateFailed` errors. When the primary chain backend (`blockstream.info`) fails to provide fee estimates or block headers during boot, the startup sequence automatically re-initializes and boots against the secondary provider (`mempool.space`) in both the foreground application and the background Notification Service Extension.

---

### Problem & Background

1. **Fatal Startup Failures on Feerate Estimation**: During `node.start()`, LDK synchronously queries the configured Esplora backend for initial fee-rate estimates (`/fee-estimates`) and the chain tip. If the primary provider is temporarily degraded, rate-limiting on carrier NATs (HTTP 429), or experiencing network dropouts, LDK raises `NodeError.FeerateEstimationUpdateFailed`.
2. **Foreground App Hang**: In `AppState.swift`, an uncaught failure during node startup aborts wallet loading, leaving the user stuck on the loading/syncing screen or requiring a manual app force-quit.
3. **Background Notification Drop**: The `NotificationService` extension operates under a strict 30-second iOS execution budget. If the primary Esplora backend stalls or fails on fee-rate estimation during node boot, background payment settlement fails and the push notification drops.

---

### Expected Behavior

- If `node.start()` fails against the primary Esplora provider, the startup routine logs `NODE_START_INITIAL_FAILED` and immediately retries startup using the secondary provider (`mempool.space`).
- If the secondary provider succeeds, the active `chainURL` is updated, `NODE_START_FAILOVER_SUCCESS` is logged, and wallet synchronization proceeds normally.
- If both providers fail, `NODE_START_FAILOVER_FAILED` is logged with the underlying error details before propagating the failure.
- In `NotificationService` (`NodeStarter.swift`), the node builder catches primary startup failure and transparently attempts a fallback node build and start using `fallbackURL`.

alienx5499:
## Summary

Adds automatic failover to the secondary Esplora backend (`mempool.space`) during LDK Node startup (`node.start()`) on iOS. Resolves fatal startup aborts caused by transient network timeouts, HTTP 429 rate-limiting, and `FeerateEstimationUpdateFailed` errors from the primary Blockstream endpoint in both the foreground app and the background Notification Service Extension.

---

## Problem & Motivation

1. **LDK Synchronous Feerate Query on Boot**: During `node.start()`, LDK Node synchronously requests feerate estimates (`/fee-estimates`) and the chain tip from the configured Esplora server.
2. **Fatal App Lockout**: When the primary Esplora endpoint (`blockstream.info`) is degraded, rate-limited, or slow to respond, `node.start()` throws `NodeError.FeerateEstimationUpdateFailed` or a network timeout. Previously, `AppState.swift` failed the startup sequence entirely, leaving the wallet in an unrecoverable loading state.
3. **Background Notification Drops**: The `NotificationService` extension has a hard 30-second operating deadline. A single failed or hung fee-rate estimation prevented the node from booting, causing inbound payments to fail settlement.

---

1. **Foreground Failover Coordinator (`AppState.swift`)**:
   - Implemented `startNodeWithFailover(mnemonic:)`.
   - Attempts initial startup using the active `chainURL`.
   - Catches startup failures, logs `NODE_START_INITIAL_FAILED`, applies a brief 300ms socket-drain backoff, and automatically re-initiates startup using the alternate provider (`mempool.space` / `blockstream.info`).
   - On failover success, updates `self.chainURL` and records `NODE_START_FAILOVER_SUCCESS`.
   - Unified all startup paths (initial launch, wallet creation, seed restore, and resume) through this resilient failover helper.

2. **Background Extension Resilience (`NodeStarter.swift`)**:
   - `DefaultNodeStarter.startNode(...)` now wraps initial node creation in a `do-catch` block.
   - If the primary builder fails or throws during `start()`, it immediately instantiates a fallback builder configured for `https://mempool.space/api` and boots the fallback node.

3. **Audit Logging & Telemetry**:
   - Emits structured audit events (`NODE_START_INITIAL_FAILED`, `NODE_START_FAILOVER_SUCCESS`, `NODE_START_FAILOVER_FAILED`) including the exact endpoints and localized error messages.

---

## Architecture Diagram

image

---

## Verification & Testing

- [x] **Primary Failure Simulation**: Verified fallback kicks in when primary URL is unreachable or returns fee estimation errors.
- [x] **Clean Startup Path**: Confirmed standard startup succeeds immediately when primary provider is healthy.
- [x] **Background Extension**: Verified `NodeStarter.swift` successfully builds and starts the fallback node within background time limits.
- [x] **Unit & Integration Tests**: All unit tests passing on Xcode destination simulator.

---

## Related 

- Closes #241


alienx5499:
> [!NOTE]
> **Marking as Draft pending PR #237 merge**
>
> To ensure the cleanest merge sequence and avoid merge churn, this PR will land after **#237** (`feat(iOS): keychain mnemonic security`).
>
> Once #237 is merged into `main`, I will rebase this branch to integrate `startNodeWithFailover` directly into the new `WalletLifecycleManager.detectStartupState()` startup cases, verify test suites, and mark this PR ready for final review.

toneloc:
Can we implement this without any dependency on #237?

toneloc:
Thanks for this. The failover idea is good, but this should not merge until we check these user-facing failure modes for accuracy:

1. **Background failover is not reliable within iOS’s notification-extension window.** The primary request may consume 15–60 seconds, and HTTP 429 responses can add roughly 16 seconds of retry backoff before the fallback begins. If iOS expires the extension during either attempt, the fallback result is discarded and the user receives **“Payment Pending — Open app to process your payment.”** This misses the feature’s main operational goal for slow or rate-limited providers. Give this flow one end-to-end deadline, fail over quickly, and reserve enough time to finish payment processing.

2. **The foreground flow releases the wallet lock between attempts.** During the 300 ms retry delay, the notification extension can take the wallet directory. The app may then wait another 35 seconds or report **“Wallet is busy,”** and two processes can briefly contend for the same LDK wallet state. Keep one lock lease across the complete failover operation so users never face a startup race or wallet-safety risk.

3. **Unrelated failures are retried and the useful error can be replaced.** The current catch-all retries database, storage, lock, entropy, and custom-LSP errors against another Esplora host even though changing hosts cannot fix them. Users can wait through two long startup attempts and then see a misleading error such as **“Node is already running”** instead of the real cause. Retry only `FeerateEstimationUpdateFailed` and `FeerateEstimationUpdateTimeout`, and preserve the original error when fallback also fails.

Before merge, verify these three outcomes with focused failure-path tests: fast background failover within the deadline, uninterrupted wallet-lock ownership, and immediate propagation of non-Esplora errors.

alienx5499:
Thanks for the feedback @toneloc! 
All requested review items have been addressed and this PR is completely standalone with zero dependency on #237. 

You can safely merge this PR (#242) first; once merged, I will immediately rebase #237 (the keychain mnemonic security PR) on top of `main` and resolve the minor conflict.

toneloc:
Pushed two commits (e78e5bf0 merge of main, e8aa972f fixes) addressing the re-review of 9132d337 — the failover design, narrow retry classification, and preserved primary error all stay; the lifecycle defects around them are fixed:

- **Wallet-lock leak (P1)**: the refcount made `release()` a decrement, so after a successful start (outer AppState lease + retained inner `NodeService.start` lease) the single release in `performBackgroundStop` left the flock held — locking the NSE out of every background stability payment while the app was suspended. `NodeDirLock` is binary again (`holdCount`/`forceRelease` deleted). The failover window is covered by **lease ownership** instead: acquire now reports whether the call newly acquired the flock, and `NodeService.start` releases on failure only when it was the actual acquirer. `startNodeWithFailover` also releases on total failure (safe no-op under binary semantics), so a broken startup can't starve the NSE either. `LSPService.switchLSP` was audited and balances unchanged.
- **NSE timeouts (P1)**: LDK's `start()` blocks on fee estimation only — wallet syncs run in background tasks afterward — so `feeRateCacheUpdateTimeoutSecs: 6` stays (that's the real failover bound and it works), while the 6s wallet-sync / 4s per-request caps are restored to 60/15. Shortening them never sped up startup, and they'd abort long-offline catch-up, holding #243's chain-freshness gate shut for exactly the wallets it protects.
- **Tests now actually run**: `NodeStartupFailoverTests.swift` is wired into `project.pbxproj` (it was absent, so CI's green never included it). The refcount tests — which required a double-release production never performs — are replaced with binary-lease tests, including the start-success → one-release regression that fails on the previous design. 7/7 pass under `xcodebuild test`.
- **Cleanups**: `isRetryableEsploraStartupError` is typed-match only (the string fallback was dead for real `NodeError`s and could false-positive on bridged errors — the old NSError test cases now assert non-retryable); the one-sided 300ms foreground backoff is removed; the working Esplora URL is written to the app group on every successful start so the NSE starts from a known-good provider; a doc comment reconciles `resolveChainURL` (launch probe) with the start-time failover.
- **Merged main** (32 commits, including #243) and reconciled `NodeStarter.swift` deliberately — gossip-strip reset reporting and the failover builder coexist.

Verified: full iOS suite (incl. the newly-wired failover tests) green on simulator, `swiftformat --lint --strict` clean with CI's 0.62.1.

PR #243, opened 2026-08-22
Prevent wallet payments before initial chain sync

Billa05:
Bug: After being offline, the wallet could reconnect and send a stability payment using its stale block height. Once chain synchronization completed, the HTLC appeared already expired, causing the wallet to force-close the channel.

Fix: The wallet must now finish its initial Lightning and on-chain synchronization before connecting to the LSP or enabling payments. If synchronization fails, startup stops safely with payments disabled.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -1617,6 +1617,18 @@ class AppState(private val context: Context) : ViewModel() {
             val amountMsat = USD(abs(result.dollarsFromPar)).toMsats(price)
             if (amountMsat == 0L) return
 
+            // Chain-freshness gate (see #243): never pay on a stale chain tip, and check
+            // BEFORE claiming so a deferral leaves no claimed-but-unsent marker. The next
+            // stability tick retries once LDK's background sync catches up.
+            val syncAge = nodeService.lightningSyncAgeSecs()
+            if (syncAge == null || syncAge > Constants.STABILITY_MAX_LIGHTNING_SYNC_AGE_SECS) {
+                AuditService.log(
+                    "STABILITY_SKIP",
+                    mapOf("reason" to "stale_lightning_sync", "sync_age_secs" to syncAge)
+                )
+                return
+            }
+
             // Atomically claim the send. A denied claim means another sender (e.g. the
             // background push service) already owns an in-flight send — skip this tick.
             val claimed = try {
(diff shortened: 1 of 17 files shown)


toneloc:
Pushed 76d12cdc to this branch (keeping your diagnosis and base commit) — it reworks the fix per the review: the startup `sync_wallets()` gate and its `process::exit(1)` are reverted, and the protection now lives at every stability-payment send boundary instead.

**Rule**: a stability payment is only sent when `latest_lightning_wallet_sync_timestamp` exists, is not in the future, and is ≤120s old (two background sync intervals).

- **Desktop**: gate inside `stable::check_stability` immediately before the send; a chain-source failure no longer prevents the app from opening.
- **Android/iOS foreground**: freshness precheck before the durable send claim + a `NodeService.sendStabilityPayment` wrapper that enforces the gate at the send itself. Deferrals never leave a claimed-but-unsent marker.
- **Android push service**: if stale, polls every 500ms within the existing 25s work budget, then defers to the foreground without claiming.
- **iOS notification extension**: if stale, polls on the background processing queue until fresh, `extensionStart`+20s, or expiry cancellation (checked under the lifecycle lock each iteration); deferral uses the existing pending-notification path, with a final backstop check in `UserToLSPHandler` before the claim.
- Gossip strips that delete `node_metrics` (resetting LDK's persisted sync timestamp) now report it, and both mobile background paths log structured `stability_gate` pilot metrics (sync age, wait time, outcome).

Verified: `cargo check --all-targets` + 184 lib tests (incl. new freshness tests), full Android `testDebugUnitTest`, iOS scheme build + `StabilityServiceTests` (42 tests incl. new freshness cases), and `swiftformat --lint --strict` all pass.

toneloc:
Pushed c26b75b5 addressing the four review findings on 76d12cdc:

- **[P1] Wait moved to the send boundary**: the iOS freshness wait now runs inside `UserToLSPHandler` after `amountMsat` is calculated — behind reconcile, cooldown, threshold, and the price fetch — so stable positions never wait or defer with a spurious "Payment Pending", and LDK's sync runs concurrently with the price work. Lifecycle safety moved with it: `NotificationService` injects a `StabilitySyncGate` (deadline, cancellation probe reading `timeExpired`/`didFinish` under the lifecycle lock, log sink) through the handler factory.
- **[P1] Deadline is authoritative**: the wait loop now checks cancellation → deadline → freshness, so a sync that becomes fresh after `extensionStart+20s` still defers and the reserved execution budget stays intact.
- **[P2] Post-claim recheck**: both mobile background paths re-check freshness immediately after the durable claim and clear the claim on staleness (claim latency under cross-process contention can cross the 120s boundary).
- **[P2] Honest metrics**: gate outcomes renamed to `fresh_ready` / `fresh_after_wait`; `sent_fresh` / `sent_after_sync` now fire only after LDK returns a payment ID, on both platforms. iOS `attempted` now also implies a payment was actually needed, so the pilot's deferral-rate denominator isn't inflated by cooldown-skip wakes.

Verified: full Android `testDebugUnitTest`, iOS scheme build (app + NSE) + `StabilityServiceTests`, and `swiftformat --lint --strict` all pass. Rust untouched by this commit.

Issue #234, opened 2026-08-12
Stable Channels — Client & Server Behavior


toneloc:
# Stable Channels — Client & Server Behavior

This is a concise description of the wallet and LSP behavior. Since the original
2026-08-31 write-up, the mobile compatibility gaps it documented have largely been
closed; the platforms now share one hardened protocol path, with the differences noted
below.

> **Version scope (2026-09-09):** Verified against `main` at [`ccd54137`](https://github.com/toneloc/stable-channels/commit/ccd54137). Major revision: the Aug-31 matrix described Android/iOS legacy behavior that has since been replaced by #248 (signed correlated trades), #273 (stabilization cap), #274/#275 (named local refusals), #252/#280 (durable splice rows), #267 (confirmation policy), and #288 (signed stability settlements, all clients).

## Core rule

Lightning finalizes the **payment**. A signed application message authorizes the
**accounting instruction** carried by that payment.

Each peer keeps its own allocation using an independently admitted BTC price. Peer
`backing_sats` is validation data, not automatically authoritative wallet state.

## Current implementation matrix

| Capability | Rust desktop | Android and iOS | LSP |
|---|---|---|---|
| `TRADE_V1` | Signs channel IDs, target USD, `trade_id`, quote and timestamp | Same hardened shape since #248 | Hardened validation for all clients |
| Trade result | Signed, durable `SYNC_V1` / `TRADE_REJECTED_V1`, correlated by trade ID, payment ID and request hash | Same since #248; results survive app kill/background (durable LSP retry + local rehydration) | Durably retries correlated responses until acknowledged; drops results only when the channel is gone from DB **and** node (#266) |
| Local trade refusals | Typed `LocalTradeAllocationError` reasons (#275) | Typed refusals with per-cause copy (#274/#275); "service unavailable" is reserved for an actually absent service | Rejection reasons are a closed signed enum — new codes require a client-first rollout |
| Stabilization cap | 99% trade-entry cap, client-side buffered maximum with Max helper (#273) | Same (#273) | `InsufficientCapacity` remains the capacity backstop |
[...]

PR #248, opened 2026-08-31
Harden Android and iOS trade-result accounting

### LINKED ISSUE #204: Update iOS and Android for signed TRADE_V1/SYNC_V1 allocations
PR #195 extended the signed mobile protocol, and PR #225 changed trade accounting so each peer preserves existing stability drift using its own locally priced delta.

Android and iOS already verify `TRADE_V1` and `SYNC_V1` signatures, but they do not yet implement all current fields and state-transition rules.

## Current compatibility paths

- Mobile `TRADE_V1` omits `quote_price` and `ts`. The LSP accepts this legacy shape, uses its own price, and allows wider fee tolerance.
- Mobile `SYNC_V1` applies only `expected_usd`; it ignores `channel_id`, `backing_sats`, and `sync_version`.

### `TRADE_V1`

- Include `channel_id`, `quote_price`, and `ts` in the signed payload.
- Validate locally before sending the non-refundable trade fee.
- Preserve existing drift by applying only the locally priced trade delta.
- Complete a pending trade only after an accepted, correlated `SYNC_V1`.

### `SYNC_V1`

- Verify the signature against the exact received payload bytes.
- Require the current `channel_id`.
- Persist `sync_version` per `user_channel_id` and reject stale or duplicate versions.
- Preserve drift using the mobile client’s own local price and delta calculation.
- Treat the LSP’s `backing_sats` as peer-state validation/audit data, not as an allocation to apply blindly.
- Update both foreground and background handlers.

## Compatibility removal

After updated Android and iOS releases are deployed and legacy usage is no longer significant:

- Require `quote_price` and `ts` for `TRADE_V1` and remove the wider legacy fee tolerance.
- Require complete, ordered `SYNC_V1` handling.

Add equivalent Android and iOS tests for both protocol flows.

toneloc:
## Summary

- Send fully correlated, signed `TRADE_V1` requests from Android and iOS.
- Persist the trade before paying its non-refundable fee; fee success alone no longer applies the trade.
- Apply only signed, channel-bound, ordered acceptances/rejections in foreground and background paths.
- Preserve the wallet’s local allocation, recover across crashes/restarts, and keep delayed results auditable.
- Pin literal SHA-256 and fee vectors across Rust, Kotlin, and Swift; add Rust pull-request CI and hardened/legacy telemetry.
- Incorporate the plan review corrections: correlated freshness is 15 minutes; legacy compatibility remains 5 minutes.

## Verification

- `./gradlew test assembleDebug` — 88 debug and 88 release tests passed; APK built.
- iOS simulator — 217 tests passed.
- Generic iOS device build — app and Notification Service Extension built successfully.
- SwiftFormat 0.62.1 strict lint passed.
- `cargo test --workspace --locked` passed.
- `cargo build --workspace --locked` passed.

## Before marking ready

- Run accepted, rejected, delayed, replay, and restart scenarios on physical Android and iPhone devices with the LSP.
- Attach the resulting audit evidence to this PR.

Closes #204

Related accuracy correction: #234

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -32,8 +32,8 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Run Rust tests
-        run: cargo test --workspace --locked
+      - name: Run affected Rust tests
+        run: cargo test -p stable-channels -p stable-channels-lsp --locked
 
-      - name: Build Rust workspace
-        run: cargo build --workspace --locked
+      - name: Build affected Rust packages
+        run: cargo build -p stable-channels -p stable-channels-lsp --locked
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -1,0 +1,39 @@
+name: Rust Tests
+
+on:
+  pull_request:
+    paths:
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "src/**"
+      - "server/**"
+      - ".github/workflows/rust-tests.yml"
+  push:
+    branches: [main]
+    paths:
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "src/**"
+      - "server/**"
+      - ".github/workflows/rust-tests.yml"
+
+concurrency:
+  group: rust-tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Run Rust tests
+        run: cargo test --workspace --locked
+
+      - name: Build Rust workspace
+        run: cargo build --workspace --locked
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -66,6 +66,10 @@ android {
         compose = true
         buildConfig = true
     }
+
+    testOptions {
+        unitTests.isIncludeAndroidResources = true
+    }
 }
 
 dependencies {
@@ -138,4 +142,5 @@ dependencies {
     // Unit tests
     testImplementation("junit:junit:4.13.2")
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.9.0")
+    testImplementation("org.robolectric:robolectric:4.16.1")
 }
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -20,6 +20,7 @@ import com.stablechannels.app.util.usdFormatted
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
 import okhttp3.OkHttpClient
 import java.util.concurrent.TimeUnit
 import okhttp3.Request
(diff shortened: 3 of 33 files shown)


toneloc:
**Rejected-scenario evidence (regtest e2e, first live run of the rejection lane) — and a bug it caught, now fixed in 85aa1ed0.**

Deterministic setup: new e2e flow `13_trade_rejected` (on `autotest`, 58ee6536) freezes the app's price refresh via a debug-only `price_refresh_secs` TestOverride while the harness moves the LSP's price 1% (limit 0.5%), guaranteeing a `quote_deviation` rejection.

**Protocol layer: correct on first contact.** Server: `TRADE_PARSED_PAYLOAD_OK` (hardened payload, trade_id + quote) → `TRADE_SIGNATURE_VALID` → `TRADE_REJECTION_QUEUED reason_code: quote_deviation` → `TRADE_RESPONSE_SENT attempt: 1`. App: signed rejection parsed, trades row persisted `status = rejected`, `expected_usd` unchanged (verified by pulling the on-device SQLite).

**The bug: both platforms' trade sheets showed “Order Confirmed ✓” for the rejected trade.** Their success check was pending-map *absence* — and a signed rejection also clears the pending map. Accounting was right; the sheet lied, with the real reason hidden in a banner behind it. Fixed in 85aa1ed0: a per-trade `TradeOutcome` (recorded only when a signed correlated acceptance/rejection commits, in both `applied` and `duplicate` branches) drives three honest sheet states — pending until a signed result, Confirmed only on acceptance, Rejected with the mapped reason.

**Re-run after the fix: 2/2 green** — flow 01 (fund) + flow 13 (reject): sheet shows “The market moved outside the quote range”, Done dismisses, allocation assert passes. Verified alongside: Android unit suite, iOS 217/217, SwiftFormat strict.

Also filed from the run: the regtest LSP can price-wedge permanently after a large mock-price step (its 10% jump guard rejects every refresh thereafter — same stuck-quarantine class as the mobile USDT-fallback issue). `make clean` clears it; worth its own harness fix.

Remaining for the ready gate: delayed / replay / restart scenarios, second physical platform, and the review's strict-keys + CI nits.

toneloc:
## Blocker resolved + scenario evidence (delayed / background / restart / replay)

**Blocker fix — `783ac700`.** The in-memory `tradeOutcomes` map was only written by the foreground control-message handler; background writers (Android `StabilityProcessingService`, iOS NSE) commit results straight to SQLite, and recovery reloaded only unresolved trades — so a trade resolved while backgrounded could stay stuck on "Order Pending". SQLite is now the source of truth the sheets can reach:

- `terminalTradeOutcome(paymentId)` (both platforms) returns `(accepted, reason_code)` for terminal rows.
- `AppState.refreshTradeOutcome` pulls terminal rows into the outcome map and clears the pending entry; swept on startup (Android) and startup + foreground (iOS).
- The Buy/Sell done screens poll the DB every 2s while the outcome is unknown, so a result committed by another process lands mid-sheet.

**Unit:** Android suite green; iOS 218/218; SwiftFormat 0.62.1 strict clean. New tests on both platforms cover rejection (persisted `reason_code`) and acceptance committed outside the handler.

**E2E (regtest harness, Maestro):** flows 01 / 13 / 14 / 15 — **all green on Android emulator and iOS simulator** (`autotest` @ `54427d64`, app code = this branch).

| Scenario | Android | iOS |
|---|---|---|
| 13 — foreground rejection (frozen quote, 1% move, 0.5% limit) | ✅ sheet shows "Order Rejected" + reason, USD unchanged | ✅ same |
| 14 — confirm → **kill app** → relaunch (accepted) | ✅ true delayed path: fee 14:45:39Z, applied **14:46:04Z post-restart**, `allocation_applied=true`, backing 4950/4950 | ✅ applied-pre-kill side of the race; $4.95 and DB row survive restart |
| 15 — confirm → **background/kill** with sheet on "Order Pending" (rejected) | ✅ node stops in background → LSP durable retry **attempts 1–3 failed, attempt 4 delivered on return**; rejection surfaced, USD unchanged | ✅ applied pre-kill; allocation + `rejected/quote_deviation` row survive relaunch |

**Replay/delivery evidence (LSP audit):** `TRADE_RESPONSE_SEND_FAILED` attempt 1–3 → `TRADE_RESPONSE_SENT` attempt 4 for the backgrounded rejection; the kill-restart acceptance delivered via LDK-level redelivery on reconnect. Duplicate redelivery is unit-covered (`DUPLICATE` → idempotent, outcome still recorded).

**Client DB dumps (post-run):** both platforms show every trade terminal — `rejected/quote_deviation` and `accepted` with correct `new_expected_usd`, no rows stuck unresolved.

Notes:
- On iOS a Maestro kill can't beat sub-second foreground delivery, so the delayed-redelivery leg is deterministically exercised on Android (backgrounding stops the node); the LSP retry worker is platform-agnostic.
- Known non-blockers left open per review: TRADE_REJECTED_V1 strict-allowed-keys forward-compat, and the CI nits (`-p sc-rest-client`, workflow comment).

PR #252, opened 2026-09-02
fix(android): graced onDestroy stop + splice-out payment tracking

DevAnuragT:
## Fixes

**1. Frequent cold node restarts on quick app open/close**
`onDestroy()` called `appState.stop()` directly, bypassing the graced stop already scheduled by `onPause()`. Since the OS can destroy/recreate the Activity on plain backgrounding (not just a real close), this forced an immediate hard stop far more often than intended, causing a resync flash on quick app-switches. Now routes through the same graced stop path as `onPause()`.

**2. Splice-out missing from History / Home screen**
Splice-out tracking relied on an in-memory flag and a DB helper hardcoded to `splice_in` only, so a restart mid-splice lost all tracking (unlike splice-in, which already persisted its row immediately). `beginSpliceOut()` now persists a `pending` row immediately, and the txid-stamping query covers both directions.

**3. Splice-out stuck showing "Failed" despite succeeding on-chain**
The one-shot LDK event that stamps the txid onto a splice-out row could be lost across a restart, and a 10-minute stale-pending cleanup would then mark the still-txid-less row as failed — even if the splice itself succeeded and confirmed on-chain. `refreshBalances()` already reads the live channel's current funding txid every call (restart-proof); it now also reconciles any stuck NULL-txid splice row against that txid and un-fails it. Verified against a real splice-out that had gotten stuck this way — the row is now correctly `completed`.

## Testing
Built, installed on a physical device, and verified against real on-chain splice-out transactions (cross-checked via blockstream.info).

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -262,12 +262,6 @@ class AppState(private val context: Context) : ViewModel() {
         paymentIds.forEach { refreshTradeOutcome(it) }
     }
     var pendingSplice: PendingSplice? = null
-    // LDK redelivers the same unhandled event until eventHandled() is called, so a sync message
-    // that can never commit (e.g. no give-up path for an uncorrelated RETRY) would otherwise
-    // wedge the entire event pipeline behind it forever. Caps retries per payment hash per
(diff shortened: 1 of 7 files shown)


toneloc:
Thanks for this. Made some fixes. Let's test again please.

PR #265, opened 2026-09-05
fix(android): ignore stale SpliceNegotiationFailed after successful negotiation

### LINKED ISSUE #263: Android: splice-out withdrawal can be shown as failed despite succeeding on-chain
## Summary

A completed splice-out (on-chain withdrawal) can be recorded as `"failed"` in Payment History even though the transaction was actually broadcast and confirmed on-chain — leaving the app's Stable USD backing overstated by the withdrawn amount.

## Observed Behavior

A splice-out withdrawal was shown as **failed** in Payment History. On-chain lookup confirms the destination actually received the funds, confirmed in block 965597 (txid `2d56a2ae9bfb08ad4bccd4f8b2e316dc080134dc593daf8cacb1146de3f085b5`) — the withdrawal genuinely succeeded. Total on-chain/lightning balance reflects the real (reduced) amount, but Stable USD does not, since it's only adjusted when the payment row completes.

## Root Cause

In `AppState.kt`'s event handler, `Event.SpliceNegotiationFailed` unconditionally reset splice state and marked the payment row `"failed"`, with no check for whether `Event.SpliceNegotiated` (success) had already fired for the same splice:

```kotlin
is Event.SpliceNegotiationFailed -> {
    val paymentRowId = pendingSplice?.paymentRowId
    isSweeping = false
    spliceTxid = null
    ...
    databaseService?.failPendingSplice(paymentRowId)
    ...
}
```

If `DatabaseService.assignPendingSpliceTxid()` fails to attach the txid to the DB row (logged as `SPLICE_TXID_UNMATCHED`) after a successful `SpliceNegotiated`, the payment row's `txid` column stays `NULL`. `failPendingSplice()`'s guard (`status = 'pending' AND txid IS NULL`) then no longer protects it, so a subsequent stale/duplicate `SpliceNegotiationFailed` event (LDK can redeliver the same queue-front event) marks a genuinely-succeeded splice as failed. Because `completeSplice()`/`StabilityService.reconcileOutgoing()` never runs, `backingSats`/`expectedUSD` are never reduced — Stable USD is left backed by more than what's actually left in the channel, and the built-in stability check can't detect this since it only compares `backingSats` vs `expectedUSD` (both stale), never against the live channel balance.

## Fix

Only treat `SpliceNegotiationFailed` as a real failure if no txid has been assigned in-memory yet (`spliceTxid == null`). If a txid is already set, `SpliceNegotiated` already succeeded for this splice, so the failure event is a stale replay and is now ignored (logged as `SPLICE_FAILED_IGNORED_STALE`) instead of rolling back state.

## Remaining Work

- Already-affected payment rows (marked `failed` despite a real confirmed splice) need a one-time manual/backend reconciliation of `backingSats`/`expectedUSD` — not covered by this fix, which only prevents new occurrences.
- cc @alienx5499 — please audit iOS's `spliceNegotiationFailed` handler in `AppState.swift` for the same race, if needed.

DevAnuragT:
Fixes #263

If `Event.SpliceNegotiated` already assigned a txid for the in-flight splice, a subsequent `Event.SpliceNegotiationFailed` for the same splice is a stale/duplicate replay (LDK can redeliver unacknowledged events). Previously this unconditionally marked the payment row `failed` and reset splice state, even though the splice had actually succeeded and confirmed on-chain — leaving Stable USD backing overstated by the withdrawn amount.

Now only treats the failure event as real if no txid has been assigned yet (`spliceTxid == null`); otherwise logs `SPLICE_FAILED_IGNORED_STALE` and leaves state untouched so the confirmation monitor completes normally.

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -143,4 +143,5 @@ dependencies {
     testImplementation("junit:junit:4.13.2")
     testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.9.0")
     testImplementation("org.robolectric:robolectric:4.16.1")
+    testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
 }
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -960,15 +960,26 @@ class AppState(private val context: Context) : ViewModel() {
                 handleSplicePending(event.channelId, event.userChannelId, "${event.newFundingTxo.txid}:${event.newFundingTxo.vout}")
             }
             is Event.SpliceNegotiationFailed -> {
-                val paymentRowId = pendingSplice?.paymentRowId
-                isSweeping = false
-                spliceTxid = null
-                spliceConfirmationJob?.cancel()
-                spliceConfirmationJob = null
-                monitoredSpliceTxid = null
-                pendingSplice = null
-                databaseService?.failPendingSplice(paymentRowId)
-                AuditService.log("SPLICE_FAILED", mapOf("channel_id" to event.channelId))
+                // If SpliceNegotiated already assigned a txid, a signed splice tx exists and
+                // may already be broadcast/confirmed (even by the counterparty). A failed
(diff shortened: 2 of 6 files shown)


toneloc (approved):
Looks good

PR #266, opened 2026-09-05
fix(android): channel-close payments getting stuck forever, and app freezing on bad sync messages

### LINKED ISSUE #264: Android: channel-close received-onchain row stuck at 0/6 confirmations due to hardcoded vout
## Summary

A channel-close "Received onchain" row in Payment History can get permanently stuck at 0/6 confirmations, even though the underlying close transaction has actually confirmed (and its funds have since been spent onward).

## Observed Behavior

After a channel close, the "Received onchain" row never progresses past 0/6 confirmed, while a later on-chain send spending those same funds already shows 1/6 confirmed. The real transactions confirmed in the correct order — this is a display bug, not a chain-ordering issue.

## Root Cause

`AppState.kt` records the channel-close payment row with `txid = null` (the close txid isn't available from the LDK close event) and relies on `CloseTxidResolver` to discover it later by polling Esplora's `/tx/{fundingTxid}/outspend/{vout}`. Both call sites hardcoded `vout = 0`. If the channel's funding output isn't at index 0, the resolver can never find the close tx, so the row's `txid` stays `null` forever — and `pollPaymentConfirmations()` skips any row with a null txid, so its confirmation count never updates.

The mempool websocket path (`WebSocketEvent.TrackedOutspend`) isn't vulnerable to this (it tracks all vouts of a txid), but it only pushes *future* events — it doesn't replay anything that happened while the app was backgrounded/disconnected. So a close that happens while the app isn't in the foreground falls back entirely on the broken REST resolver.

## Fix

Added a `fundingVout` field (mirrors the existing persisted `fundingTxid` pattern) populated everywhere the real funding outpoint is known (`refreshBalances()`, `ChannelPending` event, `prepareChannelCloseTracking()`), with a matching `closing_funding_vout` prefs fallback. Both `CloseTxidResolver.resolve()` calls now use the real vout instead of a hardcoded `0`.

## Affected Platforms

- Android (`AppState.kt`, `CloseTxidResolver.kt`)

DevAnuragT:
Fixes #264

## Problem
A channel-close payout can land on an output the app isn't tracking as its current receive address. `CloseTxidResolver` also hardcoded `vout=0` when polling Esplora, so it could only ever find a close txid if the funding output happened to be at index 0. Either way, the row was stuck at 0/6 confirmations with no txid forever.

Separately, a signed trade-sync message that could never commit locally would retry indefinitely, since LDK's event loop only advances once the current event is acknowledged. That silently blocked every later event — including `Event.ChannelClosed` — so `CloseTxidResolver` never even got a chance to run. And once a channel closed, its local `channels` row is deleted, so any trade-sync still arriving for it would be treated as a transient race and retried forever instead of being recognized as permanently invalid.

## Fix
- Added `fundingVout`, tracked the same way as the existing `fundingTxid`, and used it (instead of a hardcoded `vout=0`) everywhere `CloseTxidResolver` is called. Backfilled independently of whether `fundingTxid` changed, so existing wallets upgrading from an older version also pick it up. If the vout is still unknown (e.g. the channel already closed before the upgrade could backfill it), the resolver now skips resolving that row entirely instead of falling back to vout 0, which could otherwise attach an unrelated output's spending txid to the payment.
- Removed the amount/time heuristic txid matcher entirely. Matching by amount + timestamp proximity isn't proof of identity — mirrors what iOS already does: resolve via a direct address/output check, or leave the row unresolved rather than guess.
- Bounded how long a stuck signed trade-sync message can keep retrying (`SyncRetryTracker`), so it can no longer block the LDK event queue forever. The bound now covers the entire signed-sync processing path, including unexpected exceptions, not just the anticipated retry statuses.
- The bound is persisted to SharedPreferences rather than kept only in memory. LDK durably persists an un-acked event and redelivers it after the app process restarts, which Android can do well before 5 continuous minutes of foreground time accumulate — an in-memory-only clock reset on every restart and never actually gave up in practice. Confirmed on device: the same stuck message retried for over 4 hours across many app restarts before this fix; after it, the same message reliably gives up 5 minutes after its true first attempt regardless of how many times the app restarts in between.
- Reclassified a missing `channels` row (i.e. the channel has since closed) from `RETRY` to `INVALID` in `applyCorrelatedTradeAcceptance`/`applyUncorrelatedSyncIfNewer`, so a trade-sync for an already-closed channel gives up immediately instead of retrying in an infinite 5-minute cycle forever.
- Hardened `updatePaymentTxid()` so it only writes to a still-unresolved row and never lets two rows claim the same txid, protecting against racing close-txid resolvers.

## Verified on device
A previously-stuck 0/6 channel-close receive row, blocked behind a stuck trade-sync retry, now resolves its txid via the corrected vout and progresses through confirmations normally once the retry bound releases the event queue. Also confirmed live that a trade-sync for an already-closed channel now resolves to `TRADE_RESULT_INVALID` immediately, instead of retrying every 5 minutes forever.

Additionally verified the persisted retry bound directly against a real recurring stuck message (a counterparty resending the same unresolvable signed sync message repeatedly over several hours):
- confirmed the give-up timer survives a force-stop/relaunch mid-window rather than resetting
- confirmed `TRADE_RESULT_GIVEN_UP` fires ~5 minutes after the true first attempt, not 5 minutes after the most recent restart
- found and fixed two bugs during this testing: `SharedPreferences.apply()`'s async write could be lost if Android killed the process before it flushed, so switched to `commit()`; and the tracker was being cleared unconditionally before the DUPLICATE/APPLIED branches' own channel-reload fallback could re-defer, which wiped the persisted first-attempt on every single retry and prevented the window from ever accumulating. Clear is now only called once a branch actually succeeds.

## Update: drop permanently-invalid trade-sync messages immediately (not just from DB)
The LSP developer (Tony) reviewed the recurring stuck-message logs and pointed out the same real
message was appearing many times not because it was being resent, but because LDK re-surfaces an
un-acknowledged event on every node/app restart — one message, seen repeatedly, not several
messages. He suggested: if a sync message's channel is missing from both the local DB and the
node's live channel list, drop it immediately instead of retrying or waiting out the 5-minute
bound.

Implemented this as `deferOrDropForMissingChannel`: when the post-apply channel reload fails, it
now also checks the node's live channel list before falling back to the timed retry. If the
channel is gone from both, the message is logged as `TRADE_RESULT_CHANNEL_GONE` and dropped for
good, right away.

Verified on device against the same real stuck message: it now drops instantly on every restart
instead of waiting out the 5-minute bound each time. Note: the message still resurfaces once per
app restart (confirmed via `ldk_node.log`), which matches Tony's explanation that it's LDK
replaying the same un-acked event on startup — but since the drop is now instant, the practical
cost of each recurrence is negligible compared to before.


toneloc:
Thanks for this one.

PR #287, opened 2026-09-09
Android: fix missing biometric gate on OnChainScreen; make Payment Confirmation toggle also govern on-chain sends (iOS parity)

DevAnuragT:
## Problem

OnChainScreen (dedicated withdraw / splice-out screen) had no biometric auth gate at all, unlike SendScreen's on-chain path. Depending on which screen was used, on-chain sends were sometimes auth-gated and sometimes not.

## Fix

- Added the same auth-gate pattern already used in SendScreen to OnChainScreen's Send button.
- Made SendScreen.findActivity() internal so OnChainScreen can reuse it instead of duplicating the walk.
- Changed Payment Confirmation to a single toggle governing both on-chain and Lightning sends, matching iOS (OnChainView.swift and SendView.swift both read the same transactionAuthEnabled key, no on-chain exception).

## Important behavior change — please read

Before this PR, on-chain sends via SendScreen **always** required biometric auth, regardless of the Payment Confirmation toggle (the toggle's label only mentioned Lightning sends). After this PR, on-chain sends via both SendScreen and OnChainScreen only require auth when Payment Confirmation is ON — same as Lightning sends, matching iOS.

This means: any existing user who has Payment Confirmation turned OFF will no longer be prompted for on-chain sends from SendScreen either (previously they always were). This is intentional — it fixes an Android/iOS inconsistency and makes the toggle behavior match its (now updated) label — but it is a real reduction in the default protection for on-chain sends when the toggle is off, not just an addition of a missing gate on OnChainScreen. No migration is included to auto-enable the toggle for existing users; flagging this explicitly for maintainer awareness before merge.

## Testing

- Full unit suite passes cleanly (227 tests, 0 failures) once local untracked WIP test files are excluded.
- Updated AppAccessPreferencesTest to cover the new toggle-driven behavior for both send types.
- Built debug APK, installed on device, verified: biometric prompt appears for on-chain sends when Payment Confirmation is ON, and is skipped when OFF, for both SendScreen and OnChainScreen.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/services/AppAccessPreferences.kt
+++ b/android/app/src/main/java/com/stablechannels/app/services/AppAccessPreferences.kt
@@ -7,12 +7,11 @@ import android.content.Context
  *
  * Two independent toggles:
  * - App Unlock: requires biometric auth on launch and resume after 5s in background
- * - Payment Confirmation: requires biometric auth before Lightning sends
+ * - Payment Confirmation: requires biometric auth before sends (on-chain or Lightning)
  *
- * Auth gate rules:
- * - On-chain sends ALWAYS require auth (regardless of toggle state)
+ * Auth gate rules (matches iOS parity: a single toggle governs both send types):
  * - Seed phrase viewing ALWAYS requires auth (regardless of toggle state)
- * - Lightning sends require auth only when Payment Confirmation is enabled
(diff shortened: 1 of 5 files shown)


toneloc:
Thanks for this. This was slightly annoying so glad you fixed.

Issue #270, opened 2026-09-07
Add signed, amount-bound `STABILITY_PAYMENT_V1` settlements to Android and iOS


Billa05:
Follow-up to #235. Of the three mobile protocol gaps listed there, two were addressed by #248 ("Harden Android and iOS trade-result accounting", which closed #204): signed/correlated `TRADE_V1`/`SYNC_V1` and signed `TRADE_REJECTED_V1` with `trade_id`.

The remaining gap is stability settlements:

- Mobile clients still send the unsigned one-byte `[1]` stability marker instead of a signed, amount-bound `STABILITY_PAYMENT_V1` message.
- Confirmed on Android: both settle paths attach `byteArrayOf(1)` (`AppState.kt`, `StabilityProcessingService.kt`), and incoming handling only recognizes that marker.
- Confirmed on iOS: both settle paths attach `Data([1])` (foreground `AppState.swift:2360-2367`, NSE `UserToLSPHandler.swift:195`); receive side classifies settlements solely by the marker (`AppState.swift:1450-1453`, `StableControlParser.swift:50-52`), and `STABILITY_PAYMENT_V1` appears nowhere in the iOS tree — a signed settlement from the LSP would be dropped as unhandled control traffic.

toneloc:
Fixed the iOS notification-extension counterparty-selection gap in PR #288, commit 1734a13771f55422aa8177a64f1722d51ef5f54c.

Background settlement verification previously used the default LSP public key. A legitimate settlement from a custom LSP could therefore be recorded as an ordinary Lightning receipt and acknowledged without crediting stable backing; payment-ID deduplication would prevent a later correction through normal receipt processing.

The extension now resolves the counterparty from the LDK channel matching both stored channel_id and user_channel_id. Missing, mismatched, ambiguous, or empty peer identity returns stateUnavailable before receipt persistence, preserving the existing retry behavior. Channel readiness is not required to authenticate an already received payment. Signature, amount, direction, freshness, and replay checks remain enforced.

Regression coverage checks custom-counterparty verification and identity-resolution failures. Validation: targeted iOS settlement and database tests passed; the app and notification extension compiled; SwiftFormat 0.62.1 and git diff --check passed. This does not address the separate deferred-sync/event-queue blockage or repair already misclassified historical payments.

toneloc:
Let's close and reopen if needed.

PR #288, opened 2026-09-09
sign and verify stability settlements on all clients

Billa05:
Closes the client side of #270.

The one-byte `[1]` stability marker on TLV 13377331 is unauthenticated: any peer could keysend 1 sat with it and have the receiver treat a settlement as made (#231). This replaces it on every client with a signed `STABILITY_PAYMENT_V1` envelope on TLV 13377333.

## Sending

Android (app + background service), iOS (app + NSE) and the desktop wallet now attach only the signed record. The payload binds `settlement_id`, `channel_id`, the exact `amount_msat`, `direction`, `expected_usd`, and a 14-day created/expires window, signed with the node key. If the envelope can't be built the payment is skipped and the pending-send claim released, never sent unsigned.

## Receiving

A payment counts as a stability settlement only via a valid signed record: canonical fields, expected direction, amount equal to the actual keysend amount, channel binding, freshness with 60s skew, and a counterparty signature.

Invalid records are audited and recorded as ordinary Lightning receipts, so the sats are never lost, but backing is not credited. Replays are dropped. The `settlement_id` is burned in the same `BEGIN IMMEDIATE` transaction that credits backing, so a crash between the two can't leave backing credited with the id still replayable.

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -1538,7 +1538,10 @@ class AppState(private val context: Context) : ViewModel() {
         }
 
         val price = priceService.currentPrice.value
-        val isStabilityPayment = customRecords.any { it.typeNum == Constants.STABLE_CHANNEL_TLV_TYPE.toULong() && it.value.contentEquals(byteArrayOf(1)) }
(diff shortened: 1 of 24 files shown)


toneloc (commented):
Tested and looks good.

PR #310, opened 2026-09-11
fix(android): channel screen leaves "Closing channel..." once the clo…

toneloc:
…se completes

ChannelView read the plain appState.isChannelClosing getter and derived readiness from nodeService.channels, neither of which is observable. When a close confirmed while the screen was open, the flag cleared but nothing recomposed, so the spinner stayed up until the user left the screen or restarted the app. Collect isChannelClosingFlow and hasReadyChannel as state; both flip at close, so the screen moves to "No channel open yet".

Found by e2e flow 09 (and flow 12, which starts on that screen); both green on autotest with this change.

Claude-Session: https://claude.ai/code/session_01QaHxBaQP2kxyuRnW1mtnKK


toneloc:
AI review 4a64586fbe922993bd7cd788205907c8b42a72db

Issue #311, opened 2026-09-12
Android: splice-out completes but Stable USD balance silently fails to update, desyncing from server


DevAnuragT:
## Summary

A splice-out (on-chain withdrawal) left the app's local Stable USD balance out of sync with the server. The app showed **$19.76**, the server showed **$14.60** — a difference of about $5, which matches a $5 splice-out that had happened the day before.

## What happened (from device logs)

1. **2026-09-11 18:03:34 UTC** — a splice-out of ~$5 (6,479,000 msat, txid `cedf6d15...`) was initiated and confirmed on-chain. The payment row in the local database is marked `status=completed`, `confirmations=1`, with the txid correctly attached.
2. The app's splice-completion code (`completeConfirmedSplice()` in `AppState.kt`) marks a splice's payment row as complete, and then — as a separate step right after — recalculates and saves the updated Stable USD / backing amount to reflect the withdrawal, and finally writes an audit log entry (`SPLICE_CONFIRMED`).
3. **The payment row was marked complete, but the Stable USD/backing recalculation step and the final `SPLICE_CONFIRMED` audit log entry never happened.** The audit log has no `SPLICE_CONFIRMED` entry anywhere for this splice-out's txid, even though this entry is always written unconditionally right after the payment is marked complete.
4. Because the recalculation step never ran, the app's local Stable USD balance was never reduced to account for the withdrawal — it stayed at the pre-splice value. The server, tracking the channel independently, reflected the withdrawal correctly, producing the $19.76 vs $14.60 mismatch.

## Relevant code

`completeConfirmedSplice()` in `android/app/src/main/java/com/stablechannels/app/AppState.kt`:
- Marks the splice payment complete via `completeSplice()`.
- Then runs balance refresh + Stable USD recalculation (`StabilityService.reconcileOutgoing()`) + save to DB.
- Then logs `SPLICE_CONFIRMED`.

There is no error handling around the recalculation-and-save step. If anything in that step fails for any reason, the payment stays marked "complete," but the balance is never corrected — and nothing is logged to indicate this happened, so it's silent.

## Logs

```
payments row (id=37):
  payment_type: splice_out
  amount_usd:   4.9996181829
  status:       completed
  confirmations: 1
  txid:         cedf6d155982d24c686aeec9e1bb98189203c600f87ff91e4f3a1e785524bffb
  created_at:   2026-09-11 18:03:34 UTC

[...]

PR #303, opened 2026-09-11
fix(iOS): splice and trade sync parity

alienx5499:
## Summary

Brings iOS into 1:1 parity with Android reliability fixes from **#265**, **#266** (addressing **#268**), and **#295**. Resolves sequential LDK event loop deadlocks during trade syncs, prevents `user_channel_id` drift desyncs, eliminates false splice-out failures caused by stale replay events, and auto-clears terminal status messages after splice completion.

---

## Parity Mapping

| Android PR / Issue | Issue Solved | iOS Parity Implementation |
|---|---|---|
| **#295** (PR) | "Move confirmed" status banner lingered permanently after splice completion. | Added a 4-second auto-clear task in `AppState.completeConfirmedSplice()`. |
| **#268** (Issue in #266) | Missing channel row on closed channels wedged event retry loop indefinitely. | Keyed lookups on stable `channel_id`. Added `ChannelCloseChecking` to return terminal `.invalid` for closed channels so LDK drains the event. |
| **#266** (PR) | Uncommittable trade result syncs blocked all subsequent LDK events. | Implemented `SyncRetryTracker` (`SyncRetryTracking`) to cap retries (20 attempts or 5 min), logging `TRADE_RESULT_GIVEN_UP` and ACKing the event. |
| **#265** (PR) | Duplicate/stale `SpliceNegotiationFailed` events after broadcast falsely tore down in-flight splices. | Created `SpliceBroadcastChecker` (`SpliceBroadcastChecking`) with tri-state status (`exists`, `notFound`, `inconclusive`) and a monotonic `spliceGeneration` guard. |

---

## Key Changes

- **`SyncRetryTracker.swift` (`SyncRetryTracking`)**: Thread-safe tracker bounding retry count and duration for uncommittable sync messages with automatic key eviction.
- **`SpliceBroadcastChecker.swift` (`SpliceBroadcastChecking`)**: Queries Esplora to verify if a splice transaction exists on-chain/mempool before tearing down state.
- **`ChannelRepository.swift` (`ChannelCloseChecking`)**: Queries `channels WHERE channel_id = ?` using the primary key index. Checks `payments` and `pending_operations` to verify if a channel is closed.
- **`AppState.swift`**: Wires `spliceGeneration` counter, verified splice failure teardown, sync retry limits in `handlePaymentReceived`, and 4-second auto-clear for "Move confirmed".

---

## Related
- Closes #302 
- Android Parity: **#265**, **#266**, **#268**, **#295**

PR #314, opened 2026-09-12
fix(android): splice deduction can't be lost, reconcile is idempotent…

toneloc:
…, LSP syncs apply, rejections resurface (#311)

Four defects behind one incident (mainnet 2026-09-12: a splice-out left the app's books $5 ahead of the LSP's, and every later trade was rejected invalid_fee).

1. The splice deduction was lost to a race. Two writers completed a splice row: the splice monitor (completeConfirmedSplice → completeSplice, which matches only status='pending', then reconciles) and the confirmation poller (which completes splice rows at 1 conf and reconciles nothing). When the poller won, completeSplice matched nothing and the deduction never ran; because hasPendingSplice()/getPendingSpliceTxid() only see 'pending' rows, a restart could not repair it either. The poller usually wins on a real device — it is force-run on every mempool websocket block header, on foreground resume and whenever History is opened, while the monitor polls every 30s. Splices are now excluded from the poller: one writer owns them, and the existing status='pending' guard makes the deduction exactly-once.

2. reconcileOutgoing() / reconcileOutgoingBacking() were not idempotent. Both re-pegged backing to newExpected/price. Below par that leaves backing ABOVE the live balance, so a retry deducted a second time ($100 -> $92 -> $82), and the phantom backing masked a genuine below-par claim from the stability check. Both now preserve sats: backing is pinned to the live receiver balance, matching the LSP's backing_after_user_to_lsp_stability. A re-run sees backing <= receiver and stops — which also makes a redelivered PaymentSuccessful (the handler rethrows transient failures) safe.

3. Uncorrelated LSP syncs never applied. The lookup keyed on user_channel_id, but each side assigns its own and the LSP signs the payload with ITS id — for a JIT channel they never match, so every sync was dropped (and before #266's retry bound, redelivered forever). That is why the LSP could not correct the diverged wallet. Keyed on channel_id now, as iOS did in #303.

4. A rejection arriving while the app was backgrounded was silent. It is verified and committed by the event handler, but the only places it shows are the trade sheet and a status message set in that same moment — both die with the process, and the relaunch is a cold start. The balance simply never moved, with no explanation. The most recent failure (within the hour) now resurfaces once in the status capsule, keyed on its payment id so it cannot repeat or displace a live message.

Plus a cold-start repair for wallets already stranded: backing above the live balance is impossible, so deduct the excess once and pin backing to the balance (idempotent, no history replay). It runs off the critical startup path — inline, it delayed the first frames enough to break onboarding.

Note 1 and 3 depend on each other: a sync cannot repair books whose backing exceeds the live balance (tradeBackingAfterDelta refuses), so the repair must restore the invariant before syncs can land.

Android unit suite 280/280, including four new regression tests. Verified on the emulator against the e2e suite: canonical 10/10, the #311 repro flow (was RED) green, and flows 13/14/15/16/17/24/25/26/28/29/30 green — flow 15 included, which had been failing on defect 4.

Claude-Session: https://claude.ai/code/session_01QaHxBaQP2kxyuRnW1mtnKK

### DIFF
--- a/android/app/src/main/java/com/stablechannels/app/AppState.kt
+++ b/android/app/src/main/java/com/stablechannels/app/AppState.kt
@@ -3680,9 +3680,12 @@ class AppState(private val context: Context) : ViewModel() {
         } ?: return
         val prefs = context.getSharedPreferences(BalanceCacheKey.PREFS_NAME, Context.MODE_PRIVATE)
         if (prefs.getString(BalanceCacheKey.LAST_SHOWN_TRADE_FAILURE, null) == failure.paymentId) return
-        prefs.edit().putString(BalanceCacheKey.LAST_SHOWN_TRADE_FAILURE, failure.paymentId).apply()
+        // Mark it seen only once it is actually on screen. start() is re-invocable (ErrorView's
+        // retry button), and by then the capsule may hold a live message — recording the failure
+        // as shown there would swallow it for good, since the marker is keyed on the payment id.
         if (_statusMessage.value.isNotEmpty()) return
         _statusMessage.value = failure.outcome.message
+        prefs.edit().putString(BalanceCacheKey.LAST_SHOWN_TRADE_FAILURE, failure.paymentId).apply()
         AuditService.log("TRADE_FAILURE_RESURFACED", mapOf(
             "payment_id" to failure.paymentId,
             "resolved_at" to failure.resolvedAt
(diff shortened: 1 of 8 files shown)


toneloc:
AI review efdfc6135712e1f97c557c383fede54d9c673da3

toneloc:
AI review 78dc23d091b7c60de43f306d7740ef6087979391

toneloc:
AI review c9179c711b61cf80f492989f7b0981470e804c84

toneloc:
AI review eaec25cc19064374c5e098b372b7cebfd6c989e6

PR #315, opened 2026-09-13
fix(android): BTC unit labels, comma percentages, stuck status messages

DevAnuragT:
## Summary

Small Android fixes found while testing recent upstream changes.

- Add missing "BTC" suffix on Send/Receive amount previews and on the Home screen's balance-bar slider amounts, matching the rest of the app.
- Comma-format the price chart's percentage change (was missing thousands separators).
- Clear "Syncing wallet...", "Network unstable...", and "Finishing background sync..." status messages once the node restart actually succeeds, instead of leaving them stuck on screen. This clear is atomic (guards against a check-then-set race with a concurrent unrelated status message) and also runs on the `AlreadyRunningException` duplicate-start success path, which previously skipped it.
- Fix the close-channel flow silently swallowing a failed `closeChannel()` call in `SettingsScreen.kt` — it now resets `isChannelClosing` and shows a "Close failed" status, matching the existing (correct) behavior already in `ChannelView.kt`. Previously a failed close left the flag stuck `true` with zero trace, which made every later onchain deposit incorrectly show "Channel closing..." instead of its real state.

## Testing

Built and installed on a physical device for each change; verified no crash via logcat after each install. Reproduced the close-channel issue with a real channel close before and after the fix.


DevAnuragT:
AI review 18e3f0ce43796732dd7da78f2aed0a896e4f3fb6

alienx5499 (approved):
LGTM!

PR #325, opened 2026-09-13
fix(android): stop on-chain deposits from being dropped/miscounted during splice or close (#316)

### LINKED ISSUE #316: Android: on-chain deposit arriving during a splice-in is silently dropped, never shown
## What happened

A splice-in ("Move to Lightning") was in progress when a second, separate on-chain deposit arrived. The app never showed the new deposit anywhere — not on the Home screen, not in History — even though it was fully confirmed on the blockchain.

## Why

The app has two ways of noticing a new on-chain deposit:

1. An instant push when a transaction hits our receive address.
2. A fallback check that compares the wallet's on-chain balance to its last known value.

The instant push is disabled while a splice or channel close is in progress:

```kotlin
if (isChannelClosing || isSweeping || pendingSplice != null) { return }
```

The fallback check is disabled only while a splice is in progress (it does not skip for channel close):

```kotlin
if (currentSats > prevOnchainSats && !isSweeping && pendingSplice == null) { ... }
```

This was meant to avoid confusing a splice's own balance movement with a real new deposit. But it has a side effect: a deposit that arrives while a splice is active is dropped, not delayed — the fallback check also quietly updates its "last known balance" while skipped, so it's not simply picked up again once the splice finishes either.

Separately, the Home screen only ever queries the single latest pending on-chain receive, so even if detection worked, it can't show two pending receives at once. (History isn't affected by this particular limitation — it lists all recorded payments — but a deposit dropped by the detection issue above never gets a row to show in the first place.)

## Suggested approach

- Remove the splice/close guard on the instant push — it's tied to a specific transaction hitting our own tracked address, so it isn't at risk of confusing a splice's own balance change with a new deposit.
- The balance-delta fallback needs a proper fix for how it interacts with a splice-in that sweeps the whole on-chain balance (not just skipping the update) — this needs a bit more design thought rather than a one-line change.
- Let the Home screen track and display more than one pending on-chain deposit at a time.

## Impact

A real, confirmed on-chain deposit can become permanently invisible in the app if it happens to land while a splice-in is already running. No funds are lost — the money is safely in the wallet — but the app gives no visibility into it.

## Fix status

Implemented, builds and installs cleanly, not yet live-tested with a real splice-in + concurrent deposit:
- Instant push no longer skips during a splice/close (dedups by txid, so no double-counting).
- Balance-delta fallback now defers instead of silently erasing its baseline while skipped (full-balance-sweep edge case is still a known limitation, not fully solved).
- Home screen now tracks and shows every pending on-chain deposit, not just the latest one.
- "Move" row/button now shows its own amount, so it's clear which funds are being moved vs. which are still arriving on-chain.

DevAnuragT:
## Summary

Fixes #316 (on-chain deposits landing during a splice/close being silently dropped, misordered, or double-counted) and its follow-on regressions found while chasing it.

### What was actually wrong (issue #316 and its follow-ons)

- A deposit landing on-chain while a splice/close was in flight could be **silently dropped** — the balance-delta detector defers during that window and previously advanced its baseline anyway, losing the deposit permanently once the splice finished.
- Multiple pending on-chain deposits were shown **oldest-first with no bound**, so a pile of old stuck/never-confirming rows could push a genuinely new deposit out of the list.
- Right after a splice-in ("Move to Lightning") confirmed, **total balance could briefly double** — Lightning's channel-monitor balance already reflected the new funds while the on-chain wallet hadn't re-synced to see its input spent yet, and the two states got summed together for a few seconds.
- A splice-out that happened to pay our own tracked receive address (a self-send) could **collide** with the always-on websocket deposit recorder, corrupting either the splice's or the deposit's row.
- Two on-chain deposits arriving within the same second had **nondeterministic** display ordering.

### What changed

- `assignPendingSpliceTxid()` now reconciles a conflicting `onchain`/`received` row that shares the splice's own txid (the self-send case) instead of failing outright — candidate resolution happens **before** any delete, and the delete+update commit atomically (rollback on failure), so a legitimate deposit row can never be lost for an assignment that doesn't complete.
- Removed an earlier address-based self-send guard that would have silently dropped a genuinely different deposit reusing the same address as a pending splice-out — the websocket receive handler now records unconditionally again, as originally intended.
- The splice-confirmation monitor now syncs the on-chain wallet **once**, right when a splice's txid first confirms and before declaring it complete — fixing the balance-doubling window — without re-syncing on every 30s retry tick while completion stays deferred (e.g. no trusted price yet).
- `completeConfirmedSplice()` republishes total balance immediately from the now-synced state.
- `getPendingOnchainReceives()` selects the newest N pending rows (so new deposits are never starved out by old stuck ones) and re-sorts them oldest-first for display, with `id` as a secondary sort key so same-second ties are deterministic.
- `detectOnchainDeposit()` defers (rather than drops) during a splice/close without advancing its baseline, so a deposit landing mid-operation isn't permanently lost.
- `detectOnchainDeposit()` also skips inserting a resolved-txid deposit row if that txid is already claimed by any row — otherwise a self-send's reconciled txid could get re-inserted as a phantom duplicate once the deferred detector woke up after confirmation (found in automated review, see below).
- On splice-out completion, the detector's frozen baseline now advances by the splice's own settled amount (`AppState.advanceOnchainBaselineForCompletedSpliceOut()`), so a self-send to an **older, untracked** address no longer gets misread as a brand-new deposit — the previous txid-based fix above only covered self-sends to the *current* receive address (found in a second round of automated review, see below).
- That baseline advance is now gated on `AppState.isOwnAddress()` (backed by a new `DatabaseService.isKnownReceiveAddress()`): an **external** splice-out never raises our own on-chain balance, so applying the advance unconditionally could misattribute a concurrent, unrelated deposit's sats into the baseline instead of surfacing it. The advance also now runs *before* clearing the splice's in-memory flags rather than after, closing a race window where the periodic detector tick could see those flags already cleared but the baseline not yet caught up (found in a third round of automated review, see below).
- Extracted several pieces of this logic into pure, directly-testable functions (`AppState.shouldClearLastCloseTxid()`, `HomeScreen`'s `PendingAmountFormatter`) and added unit tests covering all of the above: splice-txid reconciliation and its rollback safety, ordering/tiebreak, close-txid clearing, and amount formatting.

### Review process

Each round of this fix was independently reviewed (adversarially, looking specifically for "fixing one thing breaks another" regressions given this exact bug's history) before moving to the next, and re-verified on a physical device after every change. An automated review caught one remaining gap: a self-send's txid, once reconciled onto the splice row, could be re-inserted as a duplicate deposit by the deferred detector after confirmation — fixed and covered by a new regression test, and confirmed on-device (fresh self-send splice-out, no duplicate row afterward). A follow-up automated review found the txid-based fix didn't cover a self-send to an older, no-longer-tracked address (unresolvable, since it has no active txid/address to match against) — fixed by advancing the detector's baseline directly by the splice's known settled amount instead, with a new pure-function regression test. A third round found the unconditional baseline advance could itself misattribute a concurrent deposit's sats when the splice-out paid an *external* address, and a smaller ordering race with the detector's periodic tick — both fixed (advance gated on address ownership, and reordered ahead of the flag clears) with new DB-layer regression tests, and re-verified with the full suite and on-device.

### On-device testing

Multiple concurrent pending on-chain deposits displaying correctly (with live $ amounts and independent confirmation counts) while a Move-to-Lightning was also in flight, with no balance doubling once it confirmed:

<table><tr>
<td></td>
<td></td>
<td></td>
<td></td>
</tr></table>

Unit tests: full suite passing (Robolectric DB-layer tests + pure-function tests), excluding one pre-existing unrelated failure in `AllowBackupTest` (manifest-classpath resource issue, not touched by this PR).


DevAnuragT:
AI review ede879838d64b20112ade666a400468de84fe946

DevAnuragT:
AI review de840841e7a30c75a9ea14444fc1f525bf579fcd

DevAnuragT:
AI review de840841e7a30c75a9ea14444fc1f525bf579fcd

DevAnuragT:
AI review cda51eebc9239e3c50d56cbe3651c92762092018

DevAnuragT:
AI review 0153e2a65f05edd2e407d4019848b618e676fb2c

toneloc (approved):

PR #326, opened 2026-09-13
Fix LSP balance-save ordering (#321)

### LINKED ISSUE #321: LSP: failed balance saves can still update memory and send SYNC
Found in the audit of `c52d43b3`. In [stable_manager.rs](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/server/stable-channels-lsp/src/stable_manager.rs#L2456), `handle_channel_ready_splice()` changes the in-memory USD target and backing before saving them to the database. If `save_channel()` fails, it logs the error but continues and can send the changed balance to the wallet. The automatic balance-repair path around line 1813 does the same.

This can leave the wallet and the running server using an update that was never saved. A later restart can reload older books. This is a code-level failure path; it has not been established as the cause of the incident in #311.

Calculate the proposed balance change, save it successfully, and only then publish it in memory and send SYNC. If saving fails, keep the previous books intact and leave the reconciliation retryable. Verify both paths with a forced database-write failure: no unsaved correction should be sent, and a later successful retry should apply the change once.

Related LSP stabilization issues: #322, #323, #324. These fixes should preserve the current Android–LSP protocol.

Billa05:
Previously, the LSP could update its running balance and send it to Android even when saving to the database failed. A restart would then bring back the older balance, leaving the books out of sync.

After the fix:

- Save first, then update the running balance and send SYNC.
- If saving fails, remember the exact correction and retry it.
- Keep receiving later events, but process their bookkeeping after the earlier correction is saved.

fixes #321


Billa05:
AI review ac664f5

Billa05:
@DevAnuragT i think this is gtg by ai review, can you take a look at this once

DevAnuragT:
> @DevAnuragT i think this is gtg by ai review, can you take a look at this once

LGTM

DevAnuragT (approved):

PR #331, opened 2026-09-14
Preserve unpaid LSP settlement when stable USD clears

### LINKED ISSUE #322: Preserve and settle amounts owed to the LSP when payments or splice-outs clear stable USD
At `c52d43b3`, payments and splice-outs can start without first settling amounts owed to the LSP. See the [payment and splice-out entry points](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/src/user.rs#L1627).

For example, a user has $10 stable USD backed by sats now worth $11. The extra $1 is still owed to the LSP, but the user spends or withdraws before paying it:

- If the spend clears the stable USD target, the remaining $1 can be incorrectly released as native BTC or left without automatic settlement.
- If a successful splice-out consumes the available user-side balance, including that $1, the value has already left the channel before the LSP receives it.

These payment and splice-out paths should require settlement before clearing the stable USD target or consuming unsettled LSP surplus. For spends that have already completed, preserve and settle the remaining obligation through reconciliation, repair, SYNC, and restart. A zero USD target alone must not release all remaining backing to native BTC.

Explicit stable USD-to-BTC trade exits already have a [settlement guard](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/src/stable.rs#L404); payments and splice-outs lack that protection. This issue covers those normal application paths, not enforcing payment against a modified or uncooperative client.

Verified in source and accounting probes at the stated baseline. No live Lightning payment or full splice-out reproduction has been run.

Related: #321, #323, #324.

Billa05:
Say you have $10 in stable USD, but BTC rises and the backing becomes worth $11. That extra $1 is owed to the LSP.

Previously, you could start a payment or splice-out before paying that $1. If it cleared your stable USD balance:

- Android could treat the remaining $1 as your native BTC.
- The LSP stopped checking settlement because your stable USD target was zero.
- A full withdrawal could take the unpaid $1 out of the channel entirely.

After the fix:

- Android blocks the payment or splice-out while an actionable stability payment is owed. It waits until settlement succeeds and the balance update finishes. You can then retry the withdrawal.
- Android preserves any unpaid backing even if stable USD reaches zero. It also prevents another spend while the previous payment’s accounting is unfinished.
- The LSP keeps checking remaining backing at zero stable USD, so clearing the balance no longer makes it ignore the outstanding settlement.

Fixes #322


Billa05:
AI review d109ced

Billa05:
AI review 435a775

Billa05:
AI review ee7f768

Billa05:
AI review 03e7bfa

Billa05:
AI review  9f87585

DevAnuragT:
Reviewed the accounting/settlement logic end to end (Android blocking, backing preservation at zero stable USD, LSP zero-target checks, and the merge conflict resolution in 9f87585). Most of it is solid and well tested. Two blockers remain, both confirmed against the code:

**1. Android: a lost LDK record for a known-ID stability marker blocks all future spends permanently**
[`OutgoingStabilityPaymentRecovery.kt#L51`](https://github.com/toneloc/stable-channels/blob/9f87585e88d84282004c39a133bd6efc057e3e90/android/app/src/main/java/com/stablechannels/app/services/OutgoingStabilityPaymentRecovery.kt#L51)
Once a pending-send marker has a `paymentId`, `node.payment(id) == null` returns `false` forever — unlike the empty-ID path, there's no age-based timeout/release here. Since `checkChannelSpend` requires `loadPendingSend() == null`, this permanently blocks all payments and splice-outs if the LDK payment record is ever lost after the marker is written. The test even codifies this as expected behavior:
[`OutgoingPaymentLifecycleTest.kt#L903-L911`](https://github.com/toneloc/stable-channels/blob/9f87585e88d84282004c39a133bd6efc057e3e90/android/app/src/test/java/com/stablechannels/app/OutgoingPaymentLifecycleTest.kt#L903-L911) (`stabilityMissingLdkRecordKeepsItsOriginalUnresolvedClaim`).
Needs a bounded recovery path for this case, same as the empty-ID branch has.

**2. Desktop: stability payment can escape classification/rollback on a DB persist failure**
[`src/stable.rs#L1161-L1177`](https://github.com/toneloc/stable-channels/blob/9f87585e88d84282004c39a133bd6efc057e3e90/src/stable.rs#L1161-L1177) sends the payment and optimistically decrements in-memory `backing_sats` before any durable record exists. The caller then persists a pending-stability row afterward:
[`src/user.rs#L1063-L1087`](https://github.com/toneloc/stable-channels/blob/9f87585e88d84282004c39a133bd6efc057e3e90/src/user.rs#L1063-L1087) — if `record_pending_stability_payment` fails, it's only logged (`STABILITY_PAYMENT_PERSIST_FAILED`), execution continues, and there's no durable marker. A later terminal event for that payment then gets misclassified as an ordinary payment (can reduce `expected_usd` incorrectly), and a failed send has nothing to roll back. This is in scope for this PR since `c6e6276a` explicitly adds the desktop-side equivalent of Android's guard ("Align the desktop guard with Android") but doesn't replicate Android's pre-send durable-claim pattern.
Needs a durable pre-send claim written before the LDK call, with the payment ID attached atomically after, matching the Android design.

Everything else checked out: Android blocking/locking across payment and splice-out paths, backing preservation at zero stable USD (no sign/off-by-one/double-debit issues), LSP-side zero-target checks (no `>0` vs `>=0` gaps), and the 9f87585 merge conflict resolution (no upstream or branch logic dropped/duplicated across the 9 conflicted files). CI (Android unit tests, Rust workspace tests, ktfmt) all pass, but neither of these two failure modes is currently covered by a test that forces the DB/LDK-record-loss path.

Billa05:
AI review c31edd2

Billa05:
@DevAnuragT can you review this, seems like ai review is not working currently

DevAnuragT:
Re-reviewed after the two new fix commits. Both original blockers are resolved as liveness bugs, but each fix replaces "stuck forever" with a new silent-accounting-loss risk when the payment actually succeeded but its LDK record was lost. Confirmed directly against code and the new tests.

**1. Android: lost-record release never debits, even after a late success proves the payment went through**
[`OutgoingStabilityPaymentRecovery.kt#L96-L107`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/android/app/src/main/java/com/stablechannels/app/services/OutgoingStabilityPaymentRecovery.kt#L96-L107) now releases the marker after a grace period, but [`DatabaseService.kt#L2079-L2090`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L2079-L2090) (`releaseLostStabilitySend`) just inserts a `failed` history row and deletes the pending-send row without any debit. If a late `PaymentSuccessful` event later confirms the payment truly went out, [`AppState.kt#L2689-L2697`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/android/app/src/main/java/com/stablechannels/app/AppState.kt#L2689-L2697) only flips the history row to `completed` — it can't call `completePendingStabilitySend` because the claim row is already gone, so the backing debit is never applied. The new test proves this: [`OutgoingPaymentLifecycleTest.kt#L933-L953`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/android/app/src/test/java/com/stablechannels/app/OutgoingPaymentLifecycleTest.kt#L933-L953) asserts `assertArchive(10.0, 11_000L)` even after the late-success event — the settlement leaves the wallet but backing/expected_usd never reflects it.

**2. Desktop: same timeout unconditionally rolls back, including the outcome-unknown case**
[`src/stable.rs#L495-L509`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/src/stable.rs#L495-L509) (`reconcile_lost_stability_claims`) treats a missing LDK record after `LOST_LDK_RECORD_TIMEOUT_SECS` the same as `PaymentStatus::Failed` and rolls back (restores backing) regardless of whether the send may actually be in flight/succeeded. If it later turns out to have succeeded, [`src/user.rs#L3449-L3457`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/src/user.rs#L3449-L3457) finds `complete_pending_stability_payment` returns `false` (already rolled back to failed) and the result is silently ignored — backing stays incorrectly restored. The existing recovery test at [`src/stable.rs#L1519-L1544`](https://github.com/toneloc/stable-channels/blob/c31edd24fe9420fcd71cc4de71385d79f324c195/src/stable.rs#L1519-L1544) only covers the genuinely-failed case, not this late-success-after-rollback case.

The pre-fix blockers (stuck app / no durable pre-send claim) are genuinely fixed — good progress. But both platforms need the "outcome unknown after timeout" case to keep the payment conservatively debited (or otherwise durably tracked) rather than reclassifying it as failed with no accounting, so a real settlement can't silently vanish from the books. CI (Android unit tests, Rust workspace tests, ktfmt) all pass, but none of these tests exercise the late-success-after-timeout-release path with a correct accounting assertion.

Billa05:
@DevAnuragT could you review now

DevAnuragT:
Re-reviewed the latest fix (head `edd1f977`, "Settle a stability payment from its success event..."). This correctly closes the two main gaps from the prior round — late success now debits on both platforms in the normal case. Two edge cases still lose the debit silently, both confirmed against code and existing tests.

**1. Desktop: debit is permanently lost if the origin channel closes before the late success arrives**
[`src/db.rs#L2937-L2981`](https://github.com/toneloc/stable-channels/blob/edd1f9771ffd98510cd3735712c6be5a32691f74/src/db.rs#L2937-L2981) (`settle_rolled_back_stability_payment`) marks the row `completed` and clears `backing_rolled_back` regardless of whether the `UPDATE channels` actually matched a row. If the channel was already closed, `changed == 0`, so `redebit` stays `None` — no error, no retry marker, transaction commits anyway. The real `ChannelClosed` handler hard-deletes the channel row via `delete_channel` at [`src/user.rs#L4489-L4506`](https://github.com/toneloc/stable-channels/blob/edd1f9771ffd98510cd3735712c6be5a32691f74/src/user.rs#L4489-L4506), even though `delete_channel` is documented as admin-only at [`src/db.rs#L1444-L1453`](https://github.com/toneloc/stable-channels/blob/edd1f9771ffd98510cd3735712c6be5a32691f74/src/db.rs#L1444-L1453) ("reconcile and channel-close paths should call `mark_channel_closed` instead"). This exact scenario is already codified (not flagged as wrong) in the new test [`src/db.rs#L5899-L5910`](https://github.com/toneloc/stable-channels/blob/edd1f9771ffd98510cd3735712c6be5a32691f74/src/db.rs#L5899-L5910), which asserts only that status becomes `completed` with no debit applied.
Fix: use `mark_channel_closed`/an archive for the origin instead of hard-deleting, and don't finalize/clear the retry marker when no debit target exists.

**2. Android: a stale archived snapshot can be debited if the channel disappears, comes back, changes, then closes again**
`deleteChannel` archives via `INSERT OR IGNORE INTO closed_channel_books` at [`DatabaseService.kt#L675-L691`](https://github.com/toneloc/stable-channels/blob/edd1f9771ffd98510cd3735712c6be5a32691f74/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L675-L691) — if an archive row for that `user_channel_id` already exists from an earlier transient disappearance, the new insert is ignored and the old (now stale) snapshot is what a later late-success settlement reads and debits against (selection/debit at `DatabaseService.kt#L2107-L2120` / `#L2179-L2186`). Existing tests only cover a single fresh archive; there's no test for archive → channel reappears/changes → closes again → late success.
Fix: upsert (replace) the archive row with the latest live allocation on each close, rather than `INSERT OR IGNORE`.

Both are narrower edge cases than the previous round's bugs (require a channel close racing with a delayed LDK success event), but same failure class — a real settlement can still silently vanish from the books. CI (Android unit tests, Rust workspace tests, ktfmt) all pass.

Still DO NOT MERGE, but very close — this is the last mile.

Billa05:
@DevAnuragT check now

DevAnuragT:
LGTM but testing on device, will approve tomorrow

DevAnuragT:
Found a new gap — confirmed directly in code.

**The released/rolled-back stability claim state isn't recognized by the balance reconcilers.** When a claim is released after a lost LDK record (no debit applied, per round 3's fix), the reconcilers that clamp backing to the live channel balance don't know about it:

- [`DatabaseService.kt#L432`](https://github.com/toneloc/stable-channels/blob/55788c0580c8d95ebf68ae6ec6413ffdecc46d22/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L432) (`clampBackingToLiveReceiver`) only checks `pending_stability_send`, never `released_stability_sends`.
- [`DatabaseService.kt#L1801-L1813`](https://github.com/toneloc/stable-channels/blob/55788c0580c8d95ebf68ae6ec6413ffdecc46d22/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L1801-L1813) (`hasPendingChannelSend`) excludes `status='failed'` rows, which is exactly what a released claim becomes.
- Same pattern on desktop: [`src/stable.rs#L157`](https://github.com/toneloc/stable-channels/blob/55788c0580c8d95ebf68ae6ec6413ffdecc46d22/src/stable.rs#L157) (`repair_overbacked_allocation_if_safe`) doesn't check `backing_rolled_back`.

Result: after a claim is released, the reconciler sees `backing_sats > live receiver balance`, treats it as a stale user overspend, clamps it down, and cuts `expected_usd` accordingly — then the next stability check pays the LSP again for the same obligation. The user can end up paying twice and having their stable balance cut twice for one $1 settlement.

Recommend a systemic fix (a single "unsettled LSP obligation" check consulted by every reconciler) rather than another point patch, given this is the same class of gap found in rounds 2-4.

DO NOT MERGE.

Billa05:
AI review 7dfba0d

Billa05:
hey @DevAnuragT 
i think this one got way too big. After several review rounds it grew to +5.6k lines, and most of that growth drifted away from what #322 actually asks for each fix for the recovery machinery's own edge cases added more machinery. Hard to review, hard to be confident in.

So I've split out the issue-specific fix as #344, a ~400-line change covering exactly what #322 describes:
   • guard payments/splice-outs that would consume unsettled LSP surplus
   • preserve residual backing when the USD target clears (instead of releasing it as native BTC)
   • let the existing stability tick / SYNC / LSP settle the residue, no new machinery

Proposing we close this PR and, once #344 lands, come back to the genuinely useful hardening that was built here as focused follow-ups

PR #332, opened 2026-09-14
LSP: retry ordinary SYNC after delivery failure

### LINKED ISSUE #323: LSP: ordinary SYNC messages are not retried after later delivery failure
At `c52d43b3`, [send_sync_message() in stable_manager.rs](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/server/stable-channels-lsp/src/stable_manager.rs#L2166) returns success as soon as LDK accepts the send request. At that point the payment carrying the balance correction can still fail. Callers only add the channel to `startup_sync_pending` when the send request itself fails.

The [PaymentFailed handler in event_loop.rs](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/server/stable-channels-lsp/src/event_loop.rs#L318) handles failed trade responses and stability payments, but does not requeue an ordinary SYNC. A wallet that needed that correction can therefore stay out of date until another event sends a new SYNC or the server restarts.

Track the delivery outcome of ordinary SYNC sends and requeue the affected channel when delivery fails, using its latest committed books and the existing signed SYNC format. Verify a send request that is accepted and later emits `PaymentFailed`: the correction should be retried and the retry cleared after successful delivery. This addresses a confirmed transport-failure gap; confirming that the wallet applied a delivered message is a separate protocol concern.

Related LSP stabilization issues: #321, #322, #324. These fixes should preserve the current Android–LSP protocol.

Billa05:
Say your stable USD balance goes from **$10 to $8** after a payment. The LSP needs to tell Android about that change.

Previously:

- The LSP started sending the balance update and treated that as success.
- If delivery failed afterward, it didn’t retry.
- Android could keep showing **$10**, while the LSP had **$8** recorded.

After the fix:

- The LSP tracks whether the update actually gets delivered.
- If delivery fails, it retries with the **latest saved balance**.
- It recovers missed delivery results after a disconnect or restart.
- Once delivery succeeds, it stops retrying.

This fix is on the **LSP side**. Android receives the same balance-update message it already understands.

Fixes #323


Billa05:
AI review b47fcab

DevAnuragT:
Verified locally — MERGE.

Ran the full test suites: root crate `db::` tests (64/64 pass) and LSP crate tests (167/167 pass), including the retry-specific ones (`sync_delivery_failure_retries_committed_books_until_delivered`, `sync_delivery_failure_survives_restart_and_reconnect_backfill`, `sync_delivery_older_outcomes_cannot_override_newer_attempts`, `sync_delivery_poll_recovers_missed_failure_and_success_events`).

Traced the logic: retry is driven by LDK's terminal payment success/failure signal, retries always reload the current committed balance (not a stale snapshot), and a monotonic `sync_version` prevents a stale retry from clobbering a newer send. Restart/reconnect recovery is durable via DB + backfill polling.

Looks correct.

DevAnuragT:
Let's wait for AI review, approved from my side

DevAnuragT (approved):

Billa05:
AI review df14dfa

Billa05:
@DevAnuragT this is gtg by ai review, plz check

DevAnuragT:
LGTM

DevAnuragT (approved):

PR #333, opened 2026-09-14
Android: apply and preserve server balance corrections

### LINKED ISSUE #318: Android: server balance corrections can be rejected or overwritten
Found while auditing Android at `c52d43b3`, after PR #314. This is a follow-up to #311: a signed server balance correction can still fail to apply, or be overwritten after it applies. This can leave the local Stable USD target out of date and contribute to later `invalid_fee` trade rejections.

In [DatabaseService.kt](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L1060), `applyUncorrelatedSyncIfNewer()` uses the capacity checks for a new trade when applying an ordinary balance correction. It returns `RETRY` for some corrections that desktop's `local_sync_backing_sats()` accepts by limiting the locally calculated backing to the available channel balance. Retrying the same state does not resolve this, and the retry window eventually expires.

Separately, [AppState.kt](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/android/app/src/main/java/com/stablechannels/app/AppState.kt#L1810) applies signed SYNC updates without `booksLock`. Meanwhile, `completeConfirmedSplice()` can save an older channel snapshot. If that save happens after a SYNC commits, it can overwrite the updated USD target and backing while keeping the newer `sync_version`. Replaying that same SYNC is then ignored as a duplicate.

Android should support the same reconciliation cases as desktop while preserving its own price and allocation rules. It should also coordinate balance writes so an older snapshot cannot overwrite an accepted correction. Verify both a correction rejected by the trade-capacity check and a SYNC arriving during splice completion.

Billa05:
This PR fixes two Android bugs when applying balance corrections from the LSP.

Previously:

- **Android could reject an existing-balance correction using checks meant for a new trade.** For example, Android still records a **$10 stable USD target**. After a payment or splice-out, the LSP sends a corrected target of **$8**, while the BTC remaining in the channel is worth **$7.99 at Android’s price**. Android could reject the correction because $8 exceeds $7.99, leaving the old $10 target recorded.

- **Splice-out completion could overwrite an accepted correction.** A splice-out moves BTC from the Lightning channel onto the Bitcoin blockchain. While that withdrawal was completing, Android could accept an LSP correction but later recalculate and save the balance using older books. In the regression test, Android had accepted **$8 stable USD**, but splice completion used the old **$10 target and backing allocation**. With **$9 worth of BTC remaining in the channel**, it incorrectly saved **$9 stable USD** over the accepted $8.

After the fix:

- Android accepts these authenticated balance corrections and calculates the backing adjustment using its own BTC price, keeping backing within the available sats.
- Splice completion uses the latest saved books, so an older copy cannot overwrite an accepted correction.
- Failed splice accounting is retried without deducting the amount twice.

This is an Android fix. The existing protocol remains unchanged: LSP corrections supply the USD target, while Android calculates its backing allocation locally. It complements #323, which fixes retrying delivery of those corrections.

Fixes #318


Billa05:
AI review 820de7c

Billa05:
the opus-5 "required" item is stale now. It was conditional on https://github.com/toneloc/stable-channels/pull/344 landing, and https://github.com/toneloc/stable-channels/pull/344 merged today with the sub-cent normalize.
i think its gtg, could you plz review it @DevAnuragT

Issue #320, opened 2026-09-13
Android: incoming stability payments can add too much to stable backing


Billa05:
At `c52d43b3`, Android adds every received stability-payment sat to stable backing in both foreground and background processing. This differs from the current Rust receive rule, which only adds the amount needed to reach the USD target at the wallet's own price and leaves any excess as native BTC.

For example, with a $10 target, BTC at $100,000, and 9,000 sats already backing the target, receiving 5,000 sats produces 14,000 backing sats on Android. The current Rust rule produces 10,000 backing sats, with the other 4,000 received sats remaining native. A focused test reproduced Android treating that extra backing as a surplus and deciding to send a stability payment back.

The affected paths are `AppState.handlePaymentReceived()` → [DatabaseService.recordPaymentAndMaybeUpdateBacking()](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/android/app/src/main/java/com/stablechannels/app/services/DatabaseService.kt#L1218) and `StabilityProcessingService.recordPaymentAtomicInDB()`. Compare them with [backing_after_lsp_to_user_stability()](https://github.com/toneloc/stable-channels/blob/c52d43b307f03c35b738959614a96f007f57bd77/src/stable.rs#L760) in `src/stable.rs`.

Both Android paths should follow the current Rust rule, including preserving backing that was already above the target before the receipt. Verify excess payments and duplicate delivery without crediting the payment twice. Related #232 proposes changing settlement policy; this issue is about Android compatibility with the current policy and message format.

PR #334, opened 2026-09-14
Android: keep excess incoming stability sats as native BTC (#320)

Billa05:
Previously:

- Android added every received sat to stable backing.
- Any excess could then look like a surplus owed back to the LSP, triggering an unnecessary payment back.

After the fix:

- Android adds only the received sats needed to fill the shortfall.
- The remaining received sats stay native BTC.
- Existing stable backing, including any surplus owed to the LSP, stays preserved.
- Duplicate events and retries cannot credit the payment twice.
- Withdrawals are still deducted correctly when receipts and withdrawals settle close together.


Billa05:
AI review 7009bd3

